PR #114, opened 2019-06-14
Finish U/G for v3.0

grantfirl:
Many updates and final touches to v3.0 User's/Technical Guide


grantfirl:
The PDF is uploaded to [https://dtcenter.org/GMTB/v3.0/scm-ccpp-guide-v3.pdf]. I'm going to go ahead and merge this since it is documentation only and does not require any retesting.

PR #120, opened 2019-06-18
Pjpegion

pjpegion:
I have added code to the SCM that allows to read in an initial condition from a FV3GFS cold start file.  The land data is also initialized, and land surface is interactive.  I also included a python script to generate the input data for the SCM from the initial conditions.  Currently, I have all of the large-scale forcing fields set to zero since I am unsure how to calculate that.  

In addition,  I made changes to gmtb_scm_vgrid.F90 to use the aks and bks that are consistent to what NCEP is using (Not GFDL's definition).  This also requires adding -DGFSL64 to CMakeLists.txt


pjpegion:
@ligiabernardet I can get started on putting your suggested changes in.  Thanks for your feedback.

pjpegion:
Quick update.  I tried to port my branch to the mac, and some of the cases fail with my branch, but work with master.  I'm looking into it.

grantfirl:
@pjpegion I was finally able to look at these changes (catching up after vacation). I have a bunch of comments/questions that I'll be adding later to this PR. I just wanted to let you know that I haven't forgotten about this!

pjpegion:
I have made my final commits.  If you comment out  -DGFSL64 in scm/src/CMakeLists.txt then the existing cases are bitewise reproducible on my mac and theia.  But I recommend leaving it in since those are the model levels the GFS is run on.  
I also cleaned up some of my changes since the gfortan compiler was not as forgiving.

grantfirl:
OK, sounds good. I'll pull in your changes and try to recompile later this afternoon after I finish up some reports to my management. Yesterday, I was still getting a compilation error on my Mac with this PR (using gfortran), so thanks for cleaning up the code to work with Mac/gfortran.

A quick related question: if I have some suggestions for code reorganization or other changes, do you want to be responsible for them or should I "take it from here"? I'm super grateful for the functionality that you've added, but I'm also cognizant that working on this model may not be in your job description! So, as I see it, there are 2 options: 1) make sure that the current PR is in a working state and accept as-is with me submitting follow-up PRs to build upon what you've coded or 2) have me write detailed comments about prospective changes that I'd like to see and have you implement them before accepting this PR. My preference is option 1. BTW, the main changes that I had in mind were to generalize the FV3 IC extraction python script, perhaps move some of the new variables around from scm_state to other DDTs, and other relatively minor things.

pjpegion:
Grant,  I prefer option 1 too.  I rather have you 'take it from here', but I will always be willing to help out if you need.  But if there is anything you see as a problem right now, I'm willing to address them.

.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = https://github.com/NCAR/ccpp-framework
 [submodule "ccpp-physics"]
 	path = ccpp/physics
-	url = https://github.com/NCAR/ccpp-physics
+	url = https://github.com/pjpegion/ccpp-physics
```
grantfirl: This change needs to be reverted. It should still point to the NCAR fork.

pjpegion: Done

scm/src/CMakeLists.txt
```diff
@@ -105,9 +106,15 @@ if (${CMAKE_BUILD_TYPE} MATCHES "Debug")
...
 elseif (${CMAKE_BUILD_TYPE} MATCHES "Release")
-  set (CMAKE_C_FLAGS       "-O2 -g -fPIC" CACHE STRING "" FORCE)
-  set (CMAKE_CXX_FLAGS     "-O2 -g -fPIC" CACHE STRING "" FORCE)
-  set (CMAKE_Fortran_FLAGS "-O2 -g -fPIC" CACHE STRING "" FORCE)
+  if(APPLE)
+    set (CMAKE_C_FLAGS       "-fmax-stack-var-size=30000 -O2 -g -I /Volumes/Cluster/opt/local/include/" CACHE STRING "" FORCE)
+    set (CMAKE_CXX_FLAGS     "-O2 -g " CACHE STRING "" FORCE)
+    set (CMAKE_Fortran_FLAGS "-O2 -fmax-stack-var-size=30000 -fbounds-check -fcheck=all -gdwarf-4 -fvar-tracking-assignments -fbacktrace -finit-real=nan -fcheck=bounds -ffree-line-length-none -cpp -fdefault-real-8 -fdec " CACHE STRING "" FORCE)
+  else()
+     set (CMAKE_C_FLAGS       "-traceback -O2 -g -fPIC" CACHE STRING "" FORCE)
```
grantfirl: Note to self for followup PR: remove -traceback option for Release build

scm/src/CMakeLists.txt
```diff
@@ -105,9 +106,15 @@ if (${CMAKE_BUILD_TYPE} MATCHES "Debug")
     message (FATAL_ERROR "This program has only been compiled with gfortran, pgf90 and ifort. If another compiler is needed, the appropriate flags must be added in ${CMAKE_SOURCE_DIR}/CMakeLists.txt")
   endif (${CMAKE_Fortran_COMPILER_ID} MATCHES "GNU")
 elseif (${CMAKE_BUILD_TYPE} MATCHES "Release")
-  set (CMAKE_C_FLAGS       "-O2 -g -fPIC" CACHE STRING "" FORCE)
-  set (CMAKE_CXX_FLAGS     "-O2 -g -fPIC" CACHE STRING "" FORCE)
-  set (CMAKE_Fortran_FLAGS "-O2 -g -fPIC" CACHE STRING "" FORCE)
+  if(APPLE)
+    set (CMAKE_C_FLAGS       "-fmax-stack-var-size=30000 -O2 -g -I /Volumes/Cluster/opt/local/include/" CACHE STRING "" FORCE)
```
grantfirl: I'm curious why these changes were needed. I believe that CMAKE_XXX_FLAGS are propagated to CMakeLists.txt for submodules, include ccpp-physics and ccpp-framework, so I'll probably want to limit using these flags for files that really need them. Also, isn't the -I include directory machine dependent? If this is necessary, we might want to only add this for a specific machine.

scm/src/gmtb_scm_forcing.f90
```diff
@@ -92,7 +92,7 @@ subroutine interpolate_forcing(scm_input, scm_state)
           v_advec_qt_bracket(1,:), top_index, 3)
 
         !>  - If the input forcing file does not reach to the model domain top, fill in values above the input forcing file domain with those from the top level.
-        if (top_index < scm_state%n_levels) then
+        if (top_index < scm_state%n_levels .AND. top_index.GT.0) then
```
grantfirl: Good catch!

ccpp/physics_namelists/input_GFS_v15_C96.nml
```diff
@@ -0,0 +1,104 @@
...
+       const_vg = .F.
+       const_vr = .F.
+       vi_max = 1.
+       vs_max = 2.
+       vg_max = 12.
+       vr_max = 12.
+       qi_lim = 1.
+       prog_ccn = .false.
+       do_qa = .false.
+       fast_sat_adj = .false.
```
grantfirl: FYI for future reference, fast_sat_adj was set to T in a previous commit. This must be set to F in the SCM because the fast saturation adjustment of the GFDL microphysics scheme is FV3-specific and cannot be run with the SCM. If it is set to T, the cloud fraction that is passed to radiation is not calculated.

scm/etc/scripts/extract_FV3GFS_column_ic.py
```diff
@@ -0,0 +1,413 @@
+#!/usr/bin/env python
```
grantfirl: This script is great, @pjpegion. It would probably have taken me a while to figure out where to find all of the necessary files to pull the appropriate data from!

I may want to generalize the interface for the script a bit, leaving the guts alone. For example, it would be nice to be able to give this script a lat/lon and for the script to find the appropriate FV3 grid point and grab that point's data. BTW, do you know of any good references for using the chgres script? I will likely need to know how to do this in the future to generate new orographic data for different resolutions.

scm/src/CMakeLists.txt
climbfuji: Any machine-dependent configurations should, where possible, go into scm/etc/MACHINE_setup.{csh,sh} and/or into doc/README_MACHINE.txt (install instructions and some system wide, default settings).

scm/src/gmtb_scm_input.f90
```diff
@@ -77,16 +80,19 @@ subroutine get_config_nml(scm_state)
   runtime = 2138400.0
   output_frequency = 600.0
   n_levels = 64
+  n_soil   = 4
```
grantfirl: Note to self: n_soil is "hard-coded" here and is not a namelist variable. This will likely need to change since the SCM could be used with the RUC LSM which uses 9 soil levels.

scm/src/gmtb_scm_type_defs.f90
```diff
@@ -69,9 +70,30 @@ module gmtb_scm_type_defs
     logical                           :: sfc_flux_spec !< flag for using specified surface fluxes instead of calling a surface scheme
     integer                           :: sfc_type !< 0: sea surface, 1: land surface, 2: sea-ice surface
     real(kind=dp)                     :: sfc_type_real(1)
+    integer                           :: veg_type !
```
grantfirl: Note to self: it may be advantageous to move these variables to a different type. The "unwritten" rule that I had been following with respect to what belongs in this data type is that it should contain variables that are either static and represent some geophysical property or that they are subject to forcing. Many of these could simply be read in to the input DDT and simply used to initialize corresponding variables in one of the other GFS DDTs (or simply variables in the physics DDT if GFS DDTs are eliminated for the SCM in the future).

grantfirl: Also, consider whether these should be array size (1) or allocatable for n_columns.

scm/src/gmtb_scm_type_defs.f90
```diff
@@ -69,9 +70,30 @@ module gmtb_scm_type_defs
...
+    real(kind=dp)                     :: snoalb(1)
+    real(kind=dp)                     :: sncovr(1)
+    real(kind=dp)                     :: uustar(1)
+    integer                           :: soil_type !< 1-14?
+    real(kind=dp)                     :: soil_type_real(1)
     integer                           :: mom_forcing_type !< 1: "revealed forcing", 2: "horizontal advective forcing", 3: "relaxation forcing"
     integer                           :: thermo_forcing_type !< 1: "revealed forcing", 2: "horizontal advective forcing", 3: "relaxation forcing"
     integer                           :: reference_profile_choice !< 1: McClatchey profile, 2: mid-latitude summer standard atmosphere
+    integer                           :: C_RES !< effective model resolution in cubed sphere resolution (needed for GWD)
+    logical                           :: model_ics !<  true means have land info too
```
grantfirl: Note to self: investigate whether this logical is truly needed; existing variables sfc_type and spec_sfc_flux might be sufficient (i.e. if sfc_type == 1 and not spec_sfc_flux, one must expect that surface ICs are provided in the case data file).

scm/src/gmtb_scm_type_defs.f90
```diff
@@ -1477,6 +1565,9 @@ subroutine scm_state_create(scm_state, n_columns, n_levels, n_time_levels)
       scm_state%v_g(n_columns, n_levels), scm_state%dT_dt_rad(n_columns, n_levels), scm_state%h_advec_thil(n_columns, n_levels), &
       scm_state%h_advec_qt(n_columns, n_levels), scm_state%v_advec_thil(n_columns, n_levels), &
       scm_state%v_advec_qt(n_columns, n_levels), scm_state%pres_surf(n_columns,1), scm_state%T_surf(n_columns,1), &
+      scm_state%alvsf(n_columns), scm_state%alnsf(n_columns), scm_state%alvwf(n_columns), scm_state%alnwf(n_columns), &
+      scm_state%facsf(n_columns), scm_state%facwf(n_columns), &
+      scm_state%hprime(n_columns,14), scm_state%stddev(n_columns), &
```
grantfirl: Note to self: investigate whether the hard-coded value of 14 can be replaced by "mntvar" variable (or whatever is used to dimension this array in the physics).

scm/src/CMakeLists.txt
pjpegion: Yes, the include file is machine dependent and should be moved out.  
But how do I tell the script to use gfortran compile flags.  CMakeLists.txt has logic for Intel, GNU, PGI comilers under teh Debug build type, but nothing for "Release". I guess instead of 'APPLE' it should be 'GNU'.  
Does this make sense?

ccpp/physics_namelists/input_GFS_v15_C96.nml
pjpegion: Understood.

scm/src/gmtb_scm_input.f90
```diff
@@ -180,22 +188,68 @@ subroutine get_case_init(scm_state, scm_input)
...
+  integer               :: input_nsoil !< number of soil levels in the input file
   integer               :: input_ntimes !< number of times represented in the input file
 
   ! dimension variables
   real(kind=dp), allocatable  :: input_pres(:) !< input file pressure levels (Pa)
   real(kind=dp), allocatable  :: input_time(:) !< input file times (seconds since the beginning of the case)
 
   !initial profile variables
-  real(kind=dp), allocatable  :: input_height(:) !< height of the pressure levels
   real(kind=dp), allocatable  :: input_thetail(:) !< ice-liquid water potential temperature profile (K)
```
grantfirl: Note to self: investigate whether ice-liquid water potential temperature should be the expected input. Temperature is more familiar to potential users and is what is available for FV3 ICs, apparently. It can also just be constructed from T, p, ql, qi. However, many cases specify their thermodynamic state in theta, theta_l, or theta_il. Maybe support both as is done here?

scm/src/gmtb_scm_input.f90
pjpegion: good point.  guess that can be added to the experiment namelist. Want me to do that?

scm/src/gmtb_scm_input.f90
```diff
@@ -180,22 +188,68 @@ subroutine get_case_init(scm_state, scm_input)
...
   integer               :: input_nlev !< number of levels in the input file
+  integer               :: input_nsoil !< number of soil levels in the input file
   integer               :: input_ntimes !< number of times represented in the input file
 
   ! dimension variables
   real(kind=dp), allocatable  :: input_pres(:) !< input file pressure levels (Pa)
   real(kind=dp), allocatable  :: input_time(:) !< input file times (seconds since the beginning of the case)
 
   !initial profile variables
-  real(kind=dp), allocatable  :: input_height(:) !< height of the pressure levels
```
grantfirl: Note to self: was I ever using this information for anything?

scm/src/CMakeLists.txt
climbfuji: Yes, you can configure different flags for different compilers in the same way as this is done for the Debug build type. Things like `-fmax-stack-var-size=30000` should go into the machine setup script.

scm/src/gmtb_scm_type_defs.f90
```diff
@@ -1509,6 +1608,22 @@ subroutine scm_state_create(scm_state, n_columns, n_levels, n_time_levels)
 
   end subroutine scm_state_create
 
+  subroutine scm_input_create_model_ics(scm_input, nlev_soil,nlev)
```
grantfirl: @pjpegion I'm curious why a separate subroutine is needed for this.

scm/src/gmtb_scm_setup.f90
```diff
@@ -42,72 +42,131 @@ subroutine set_state(scm_input, scm_reference, scm_state)
...
+        scm_state%snoalb = scm_input%input_snoalb  
+        scm_state%sncovr = scm_input%input_sncovr  
+        scm_state%area = scm_input%input_area    
+        scm_state%tg3    = scm_input%input_tg3     
+        scm_state%uustar = scm_input%input_uustar  
+        scm_state%soil_type_real = DBLE(scm_input%input_soiltyp)
+        scm_state%stc(i,1,:,1)=scm_input%input_stc
+        scm_state%smc(i,1,:,1)=scm_input%input_smc
+        scm_state%slc(i,1,:,1)=scm_input%input_slc
+        if (scm_input%input_pres_i(1).GT. 0.0) then ! pressure are read in, overwrite values
```
grantfirl: Is this check for input_pres_i > 0 really needed? Can we assume that if we have FV3GFS ICs, that pres_i, and pres_l will just be provided already?

scm/src/gmtb_scm_type_defs.f90
```diff
@@ -1648,6 +1795,27 @@ subroutine physics_associate(physics, scm_state, col)
     physics%Sfcprop(col)%tsfc => scm_state%T_surf(col,:)
     physics%Sfcprop(col)%tref => scm_state%T_surf(col,:)
     physics%Sfcprop(col)%slmsk => scm_state%sfc_type_real
+    physics%Sfcprop(col)%vtype => scm_state%veg_type_real
+    physics%Sfcprop(col)%vfrac => scm_state%veg_frac_real
+    physics%Sfcprop(col)%slope => scm_state%slopetype
+    physics%Interstitial(col)%sigmaf = min(scm_state%veg_frac_real,0.01)
```
grantfirl: Note to self: Check to see if this is working as expected. The physics%Interstitial DDT gets reinitialized every time step, so this value might be getting cleared out before getting used.

scm/src/gmtb_scm_type_defs.f90
```diff
@@ -1679,6 +1847,16 @@ subroutine physics_associate(physics, scm_state, col)
       physics%Sfcprop(col)%spec_lh_flux => scm_state%lh_flux
       physics%Sfcprop(col)%zorl => scm_state%sfc_roughness_length_cm
     endif
+    if(scm_state%model_ics) then
```
grantfirl: I wonder why all surface-related variables aren't wrapped in the model_ics test?

scm/src/gmtb_scm_setup.f90
pjpegion: Actually, the extract_FV3GFS_column_ic.py script only pulls out surface pressure and the SCM uses the aks and bks to calculate pres_i and pres_l.  I want this functionality there for more through debugging of a model crash. Since the model full model is non-hydrostatic, the aks and bks do not give the exact pressure after the initial time.  I have a different python script that can pull out more information from the history files of a run to diagnose a crash.

scm/src/gmtb_scm_type_defs.f90
pjpegion: These surface fields were already there, so I did not want to break existing functionality, and I the surface_roughness_lenght is also used by the MP and PBL schemes.

scm/src/gmtb_scm_type_defs.f90
pjpegion: Guess It does not,  I was just trying to keep the additional allocations separate.  It could be added to the scm_input_create subroute and bracketed by an if statement.

scm/src/gmtb_scm_input.f90
pjpegion: Actually, the FV3 initial conditions only have geopotential, and I calculate temperature from that.
I initially tried to calculate ice-liquid water potential temperature, but I couldn't reproduce the full model's physics tendencies, so I switched to temperature.

scm/src/gmtb_scm_input.f90
grantfirl: No, I can do that in a followup PR. It might be nice to do something fancier too like autopopulate its value based on the land surface scheme in the CCPP SDF.

scm/src/gmtb_scm_setup.f90
grantfirl: OK, thanks for the explanation.

scm/src/gmtb_scm_type_defs.f90
grantfirl: OK, thanks.

scm/src/gmtb_scm_input.f90
climbfuji: NEMSfv3gfs uses Model%lsm, Model%lsm_{noah,noahmp,ruc} = {1,2,3} for this purpose.

scm/etc/scripts/extract_FV3GFS_column_ic.py
pjpegion: @grantfirl  The chgres script I'm familiar with only interpolates the spectral GFS initial conditions to the cubed-sphere grid.  I don't know which script creates the boundary condition files for a given resolution.  I recall Jeff Whitaker asking EMC to create these files when we wanted to run a different resolution than what was already available.

grantfirl:
See https://github.com/pjpegion/gmtb-scm/pull/1. Once that PR is accepted into pjpegion's fork, this PR should be updated to target NCAR/v3.0 rather than NCAR/master. Once that is done, this should be ready to accept and merge. With the changes in https://github.com/pjpegion/gmtb-scm/pull/1, I've tested the new FV3GFS point ICs on Mac/GNU, Cheyenne/Intel, Theia/Intel.

grantfirl:
I made a mistake by putting a PR into prpegion/v3.0 rather than prpegion/prpegion for https://github.com/pjpegion/gmtb-scm/pull/1. See https://github.com/pjpegion/gmtb-scm/pull/2 now. Once that is merged, and the target of this PR is changed to NCAR/v3.0, we should be in business.

climbfuji:
As far as I can see this affects gmtb-scm only, in which case I am a little more relaxed ;-) No, seriously: I looked at the changes, the only two concerns I have are that we don't really want personalized setup scripts in scm/etc (those should stay in personal branches imho), and that the size of the C96 input files is somewhat at the edge of what GitHub allows and what users teleworking can suffer. If those are getting more or larger, we will need to find a place and method to store them (remember the discussion about setting up sandboxes for NEMSfv3gfs using gitlfs, gitlab, ...). I'll go ahead and approve.

grantfirl:
Agreed. I forgot to mention that @pjpegion should take the machine setup script and put it in his own forks/branches so that we can remove it from the NCAR fork. As far as the input files, I also agree that we should not be relying on the repo to house these kinds of things in the future. What is in here now is a (very helpful!) proof-of-concept, with followup PRs to address several comments raised in this PR. I have plans to generalize the FV3GFS_ICs-to-SCM_ICs script so that a user can point to whatever external data they have (that will be housed outside of the repo) and to include advective forcing from the UFS. Once again, sincere thanks to @pjpegion for pushing the envelope and getting this process started.

PR #121, opened 2019-07-03
Update cheyenne modules

llpcarson:
Update v3.0 release branch with new cheyenne modules following cheyenne o/s upgrade.
The multi-suite test runs successfully for Intel and Gnu on cheyenne. PGI currently fails to compile.

### DIFF
--- a/scm/etc/Cheyenne_setup_gnu.csh
+++ b/scm/etc/Cheyenne_setup_gnu.csh
@@ -5,11 +5,11 @@ echo "Setting environment variables for SCM-CCPP on Cheyenne with gcc/gfortran"
 #load the modules in order to compile the GMTB SCM
 echo "Loading gnu and netcdf modules..."
 module purge
-module load ncarenv/1.2
-module load gnu/8.1.0
-module load ncarcompilers/0.4.1
+module load ncarenv/1.3
+module load gnu/8.3.0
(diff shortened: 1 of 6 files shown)


climbfuji:
Have you tested those changes? Does the model compile and run the mother-of-all-run-scripts?

llpcarson:
I only ran one case of the SCM.  Will run the big-test now...

climbfuji:
We should do this with all three compilers ... thanks!

llpcarson:
The multi-suite test runs successfully for Intel and Gnu on cheyenne.  PGI currently fails to compile.

grantfirl:
Since @climbfuji asked in an email, the PGI failure is as follows:

`PGF90-F-0000-Internal compiler error. mark_ilitree: visit != val      -1  (/glade/u/home/grantf/gmtb-scm-v3.0/ccpp/physics/physics/cu_gf_deep.F90: 3896)
PGF90/x86-64 Linux 19.3-0: compilation aborted`

I have no idea how to go about troubleshooting this. The line where it is failing is in a code stanza that is updating a couple local variables. Actually, this code stanza doesn't make sense to me, since it is updating variables that are local to the subroutine right before the end, so the values aren't even used. If these lines are commented out, the compiler just seg faults with signal 11.

PR #127, opened 2019-09-13
Update SCM to work with gmtb/develop of CCPP, 2019/09/13

climbfuji:
- Update to new metadata format
- changes for GSD suite from CCPP's gsd/develop branches
- latest changes in IPD physics from Vlab master
- CIRES UGWD physics
- contains several commits from @grantfirl 

All 20 runs complete on macOS with LLVM clang and GNU gfortran.

Needs to be tested for b4b identical results (except for GSD physics, which has changed).

Documentation needs to be updated to reflect new metadata format.


climbfuji:
Awesome, thanks for testing. I'll leave it up to you then to approve and merge if the changes look good to you (and add missing changes to the documentation?)

scm/src/GFS_typedefs.F90
```diff
@@ -2901,8 +2989,16 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
     Model%cny              = cny
     Model%lonr             = gnx         ! number longitudinal points
     Model%latr             = gny         ! number of latitudinal points from pole to pole
-    allocate(Model%blksz(1:size(blksz)))
+    ! DH*
+    write(0,*) "DH DEBUG GFS_typedefs.F90: size(blksz)=", size(blksz)
```
grantfirl: @climbfuji Need to leave these debug statements in?

scm/src/GFS_typedefs.F90
```diff
@@ -2901,8 +2989,16 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
...
+    write(0,*) "DH DEBUG GFS_typedefs.F90: blksz=", blksz
+    call sleep(5)
+    ! *DH
+    Model%nblks            = size(blksz)
+    allocate(Model%blksz(1:Model%nblks))
     Model%blksz            = blksz
+    allocate(Model%blksz2(1:Model%nblks))
```
grantfirl: Maybe I don't need to understand this, but why are there two blksz vars? Is this related to the FV3's non-uniform blocks thing?

climbfuji: We need two different ones for horizontal_loop_extent and horizontal_dimension. In FV3, those had changes from Interstitial%im and Interstitial%ix to Model%blksz(blkno) and Model%blksz2(blkno), and I made the same changes for SCM. I realize this is not necessary. Even if we have multiple columns ("blocks"), then each block will have the same length of one. Should we fix this in this PR or in a follow-up PR?

scm/src/GFS_typedefs.F90
climbfuji: No. Will remove them, thanks for catching this!

scm/src/GFS_typedefs.F90
grantfirl: OK, thanks for the explanation. I think that it's OK to leave it in to try to reduce the number of differences between the FV3 and SCM versions of GFS_typedefs. There are many instances when something is not needed for SCM but it is left in for this reason.

scm/src/GFS_typedefs.F90
climbfuji: Ok, I removed those debug prints.

PR #131, opened 2019-09-18
pre-compile/compile NoahMP for SCM

grantfirl:
update to allow compilation with NoahMP code; added SDF and associated namelists for using NoahMP

Note that this will not run yet since modifications will need to be made to initialize NoahMP-specific variables correctly (following what pjpegion did for Noah)


grantfirl:
This PR will also need to be updated to include @climbfuji changes to the NoahMP PR.

grantfirl:
Associated PR:
https://github.com/NCAR/ccpp-physics/pull/321

grantfirl:
> This PR will also need to be updated to include @climbfuji changes to the FV3 NoahMP PR, but this passes the ccpp_prebuild and compilation stages for the SCM as-is.

This PR was updated to include @climbfuji's changes to GFS_typedefs in FV3. This pre-compiles/compiles so that it can be used with the gmtb/develop branches of ccpp-physics (but doesn't run due to initialization).

climbfuji:
> > This PR will also need to be updated to include @climbfuji changes to the FV3 NoahMP PR, but this passes the ccpp_prebuild and compilation stages for the SCM as-is.
> 
> This PR was updated to include @climbfuji's changes to GFS_typedefs in FV3. This pre-compiles/compiles so that it can be used with the gmtb/develop branches of ccpp-physics (but doesn't run due to initialization).

@grantfirl I am sorry I don't understand really what the issue is. Can't the initialization of NoahMP/soilveg for SCM happen in GFS_phys_time_vary.scm.F90? There were no changes in GFS_typedefs required for initializing set_soilveg in GFS_physics_driver.F90?

grantfirl:
> Thanks for the explanation! We could maybe add a guard to prevent the user running a suite that includes NoahMP, but I leave it up to you.

OK, I added a stop when NoahMP is chosen in control_initialize.

PR #133, opened 2019-10-09
generalized UFS IC generator + NoahMP running

grantfirl:
This PR can be a considered a followup PR for both #120 and #131 . In a nutshell, these changes generalize the previous script to generate SCM ICs from UFS Atmosphere ICs. This script can now be used to generate both Noah and NoahMP ICs appropriate for the SCM from cold start UFS ICs. Using new ICs generated by this script, the SCM can now run on top of interactive Noah or NoahMP LSMs for any point in the (supplied) UFS atmosphere grid. It has been briefly tested using the default GFS_v15 suite (with Noah) and a modified suite (with NoahMP).

To use the new script (in scm/etc/scripts/UFS_IC_generator.py):
usage: ./UFS_IC_generator.py [-h] (-l LOCATION LOCATION | -ij INDEX INDEX) -d
                           DATE -i IN_DIR -g GRID_DIR [-t {1,2,3,4,5,6}]
                           [-a AREA] [-mp] -n CASE_NAME

Mandatory arguments: 
1. EITHER (lon/lat pair in decimal degrees) or (UFS grid i,j index with the tile number)
2. date corresponding to the ICs
3. directory containing the UFS ICs
4. directory containing the UFS supergrid files (AKA "fix" directory)
5. case name (what to call the output netCDF file)

Optional arguments are:
1. tile (if one already knows which tile the desired lon/lat is contained in or if one is supplying (i,j) indices)
2. whether or not to generate NoahMP ICs in the output file

To test this functionality, example FV3 ICs were added to the repo. I ran the following commands:

From etc/scripts:
`./UFS_IC_generator.py -l 261.51 38.2 -d 201610030000 -i ../../data/raw_case_input/FV3_C96_example_ICs -g ../../data/raw_case_input/FV3_C96_example_ICs -n fv3_model_point_noah`

This generates SCM ICs for a point in central KS for 3 Oct 2016 on the C96 FV3 grid. Note that the IN_DIR (-i) and GRID_DIR (-g) arguments are the same in this case (since I copied the supergrid files to the same directory as the IC files for point of example), but they will not in general be the same if you're given UFS ICs.

A similar command is used to generate NoahMP ICs (just use the -mp argument):
`./UFS_IC_generator.py -l 261.51 38.2 -d 201610030000 -i ../../data/raw_case_input/FV3_C96_example_ICs -g ../../data/raw_case_input/FV3_C96_example_ICs -n fv3_model_point_noahmp -mp`

Following the generation of ICs, run the SCM as follows for two new FV3 point cases that have been added (from bin):
`./run_gmtb_scm.py -c fv3_model_point_noah -s SCM_GFS_v15 -n input_GFS_v15.nml`
`./run_gmtb_scm.py -c fv3_model_point_noahmp -s SCM_GFS_v15_noahmp -n input_GFS_v15_noahmp.nml`

Finally, run gmtb_scm_analysis.py using plot_configs/noahmp_test.ini to plot some basic things (from etc/scripts):
`./gmtb_scm_analysis.py plot_configs/noahmp_test.ini`


grantfirl:
More Notes:

- fixed bug in ccpp_prebuild_config.py (wrong group for NoahMP)
- remove NoahMP stop statement in GFS_typedefs.F90
- remove lon/lat time dependency (column is assumed to be stationary for simplicity)
- add NoahMP var processing in gmtb_scm_input
- add (n_columns,im) dimensions to Noah and NoahMP vars
- add NoahMP vars to input and state types
- lat/lon is now expected in scalar group rather than forcing (in case data file)
- add NoahMP vars to physics%associate
- add physics namelist and SDF for GFS_v15_noahmp
- add processed case files and case configuration files for example fv3_model_point_noah and fv3_model_point_noahmp cases
- add example UFS ICs directory
- add (original) NoahMP tables (not the Fortran-ized version used in the UFS)
- regenerate all existing processed_case_input files (and modify generation python scripts) to account for now-stationary column

grantfirl:
NOTE: This work was done before ccpp/physics was updated to have the latest GFS v16 changes. So, this was tested with commit d808b1 of ccpp-physics (https://github.com/NCAR/ccpp-physics/commit/d808b16ec1bda4cdc0c99e88877cd7a19f2b9329)

scm/etc/scripts/UFS_IC_generator.py
```diff
@@ -0,0 +1,1489 @@
...
+            message = 'The latitude {0} is outside of the range {1}'.format(location[1], '[-90,90]')
+            logging.critical(message)
+            raise Exception(message)
+    
+    date_dict = {}
+    if len(date) != 12:
+        message = 'The entered date {0} does not have the 12 characters expected in the format YYYYMMDDHHMM'.format(date)
+        logging.critical(message)
+        raise Exception(message)
+    else:
```
climbfuji: I believe the Pythonic way to do this is use `dt = datetime.datetime.strptime(date, '%Y%m%d%H%M')` and then assign `dt.year`,... to `date_dict` (if the latter is needed at all, `dt` would probably do)

grantfirl: Thanks. That would def be much cleaner.

grantfirl: You betcha, lots of changes, which is why I wanted to merge a chunk in before continuing on. I definitely hear you about the hard-coded stuff. Many of the existing scripts that I had to change were because of hard-coding done in the past, which was done due to expediency and the fact that most of those scripts were basically run-once (unless the netCDF format that the SCM expects changes, which it has only been once or twice in 3 years). I agree though, that the UFS_IC_generator.py script (which should be used frequently in the future) should be as free of hard-coding as possible, which can be a goal of future clean-up. I think that I have taken the key user-changeable variables and put them as script arguments, though, which I hope can be considered an improvement over the original proof-of-concept script.

PR #140, opened 2019-11-07
Doc update 20191106

grantfirl:
Updates to the User/Technical Guide to reflect the added static option and UFS IC capability.


scm/doc/TechGuide/chap_quick.tex
```diff
@@ -109,19 +109,22 @@ \section{Compiling SCM with CCPP}
...
 \begin{lstlisting}[language=bash]
 ./ccpp/framework/scripts/ccpp_prebuild.py --config=./ccpp/config/ccpp_prebuild_config.py [--debug]
 \end{lstlisting}
+A second, more performant ``static'' mode is available that generates software caps for each physics group defined in the supplied SDFs and generates a static library that becomes part of the SCM executable. Using this mode, flexibility of defining suites at runtime is lost, although one can choose among a set of suites defined at compilation time. To use this mode, the syntax becomes
+\begin{lstlisting}[language=bash]
+./ccpp/framework/scripts/ccpp_prebuild.py --config=./ccpp/config/ccpp_prebuild_config.py --static --suites=suite list [--debug]
+\end{lstlisting}
+where the supplied suite list is a comma-separated list of suite names that exist in the ./ccpp/suites directory.
```
climbfuji: I think we need to say here that the suite names are the suite definition files minus the suite_ prefix and minus the .xml suffix. Maybe add a line showing an example for three suites? This reminds me that we want to have a better solution for building the code, which involves running ccpp_prebuild.py as part of the cmake command. For the static build, all suites in the ccpp/suites directory should be compiled into the executable so that the multi-run script works.

grantfirl: Great suggestion, thanks. I added three example suite names and mentioned what you suggested.

PR #149, opened 2019-12-03
add files to run Ferrier-Aligo MP scheme with SCM

grantfirl:
added SDF and namelist to run with Ferrier-Aligo MP + added initialization data files for same


ccpp/physics_namelists/input_GFS_v15_FA.nml
```diff
@@ -0,0 +1,139 @@
...
+       xkzminv        = 0.3
+       xkzm_m         = 1.0
+       xkzm_h         = 1.0
+       do_sppt        = .false.
+       do_shum        = .false.
+       do_skeb        = .false.
+       do_sfcperts    = .false.
+/
+
+&gfdl_cloud_microphysics_nml
```
climbfuji: Just for curiosity, do we need this section for FA?

grantfirl: Not that I know of. This was just a copy/paste thing.

PR #150, opened 2019-12-05
Fix noahmp precip rates

grantfirl:
Since NoahMP precipitation rate calculation was moved to GFS_MP_generic_post_run, sfc_noahmp_pre is no longer needed, so it is deleted from the CCPP configuration file and SDFs. Also, precipitation rate variables used in RUC LSM and NoahMP LSM are put in the Sfcprop DDT rather than the catch-all TBD DDT.


grantfirl:
Associated PRs:
https://github.com/NCAR/fv3atm/pull/14
https://github.com/NCAR/ccpp-physics/pull/365

PR #154, opened 2019-12-11
CCPP/SCM short course prep

grantfirl:
renamed GFS_v15.2 files to GFS_v15p2 (since the . character is not allowable in Fortran subroutine names); added prescribed surface versions of GFS_v15.2 and GFS_v16beta; added configuration file for running multi_run_gmtb_scm.py for short course; added plot configuration files for basic twpice and LASSO plots; other minor plotting changes


climbfuji:
Good point about the . - for consistency with how we are substituting . in standard_name, it would be better to use a lowercase p instead of _ (the _ is a substitute for -).

grantfirl:
> Good point about the . - for consistency with how we are substituting . in standard_name, it would be better to use a lowercase p instead of _ (the _ is a substitute for -).

Thanks. Done.

grantfirl:
Added user-configurable option for plot file type in gmtb_scm_analysis.py ('plot_ext' variable at the top). On my machine, available default file types include pdf, ps, eps, and png. PNG runs the fastest and is probably OK for most purposes, but one of the vector formats is better for presentations/publications.

grantfirl:
Added statements to run_gmtb_scm.py to link plotting scripts and configuration files to the run directory per email communication.

grantfirl:
@climbfuji @llpcarson This PR is actually needed to fix dtc/develop. The 15.2 naming error and another Ferrier-Aligo SDF name error prevent this from compiling with the static build (missed because I tested the previous PRs only using the dynamic build). I'd like to merge right away, unless you have time to review.

scm/etc/scripts/plot_configs/LASSO_2016051812_force_compare.ini
```diff
@@ -1,7 +1,7 @@
-gmtb_scm_datasets = ../../bin/output_LASSO_2016051812/output.nc, ../../bin/output_LASSO_2016051812_VARA/output.nc, ../../bin/output_LASSO_2016051812_MSDA/output.nc,
+gmtb_scm_datasets = output_LASSO_2016051812/output.nc, output_LASSO_2016051812_VARA/output.nc, output_LASSO_2016051812_MSDA/output.nc,
```
climbfuji: I like this MUCH better

PR #156, opened 2020-01-07
Fix Hera plotting

grantfirl:
Hera uses an older version of the Pandas package where the following line causes an error:
`pandas.plotting.register_matplotlib_converters()`

This line was added because not having it there with newer versions of Pandas leads to a "Future Warning". This was put in a try block.

Also, it appears that Hera doesn't have dvipng installed, which is needed to render LaTeX in plots and save as a PNG file. The file format has been changed back to using PDF instead.

Note: This should be merged after the short course since instructions mention PNG, and the AMI has no problem with plotting PNGs.

### DIFF
--- a/scm/etc/scripts/gmtb_scm_analysis.py
+++ b/scm/etc/scripts/gmtb_scm_analysis.py
@@ -18,12 +18,15 @@ Rd = 287.0
 Rv = 461.0
 g = 9.81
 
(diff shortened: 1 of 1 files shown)


grantfirl:
@ligiabernardet This should fix python plotting on Hera for after the short course.

ligiabernardet:
Thank you!

PR #166, opened 2020-03-09
Various updates to the TechGuide. 

ligiabernardet:
Removed mention of GMTB from acknowledgements. Changed an example from an unsupported suite to a supported one. Defined UFS before first occurrence. Updated links to CCPP Tech Doc v4 (note that these are broken links until the CCPP TechDoc v4 is placed in its final location. Corrected that SDF is read at compile, not run time (since only static build is supported). PDF was successfully built.


ligiabernardet:
In addition to the proposed changes in this pull request, I have a couple of questions/comments on the document.
1. Should the cloning instructions point to a tag (not branch)?
2. There is substantial mention of Noah MP but it is not part of any supported suite. Should it be clarified that this is not a supported capability? In the future we can add a suite with NoahMP to the list of our supported suites.

ligiabernardet:
The CCPP v4 TechDoc is now in its final location so the link I updated is now working fine (no more known broken links).

grantfirl:
> In addition to the proposed changes in this pull request, I have a couple of questions/comments on the document.
> 
> 1. Should the cloning instructions point to a tag (not branch)?
> 2. There is substantial mention of Noah MP but it is not part of any supported suite. Should it be clarified that this is not a supported capability? In the future we can add a suite with NoahMP to the list of our supported suites.

Yes, I suppose it should point to a tag and not a branch. I can change that.

I can add something that says NoahMP is not supported, but you can work with it at your own risk.

PR #168, opened 2020-03-09
Machine setup script change

grantfirl:
change generic machine setup scripts to check for environment variables and provide instructions if not set


grantfirl:
OK, suggested changes completed.

climbfuji:
> OK, suggested changes completed.

Thanks! I am not sure how much we want to change given the tight deadlines. But the current setup scripts for macOS, UBUNTU and CENTOS are not fully compatible with the UFS install instructions in NCEPLIBS-external/doc.README_*.txt. I don't know whether we need to be, though, but at least the scripts should work with both the NCEPLIBS-external instructions as well as the instructions in the SCM User's Guide. What do you think?

grantfirl:
> > OK, suggested changes completed.
> 
> Thanks! I am not sure how much we want to change given the tight deadlines. But the current setup scripts for macOS, UBUNTU and CENTOS are not fully compatible with the UFS install instructions in NCEPLIBS-external/doc.README_*.txt. I don't know whether we need to be, though, but at least the scripts should work with both the NCEPLIBS-external instructions as well as the instructions in the SCM User's Guide. What do you think?

It would be nice to be as compatible with the UFS release as possible. Please let me know how  they are not compatible, and I will work to make them so.

climbfuji:
See https://github.com/grantfirl/gmtb-scm/pull/2, please!

PR #170, opened 2020-03-10
Address testing failures for UFS_IC_generator.py

grantfirl:
- obtain grid center lon/lat values from supergrid tile data (CXXX_grid.tileX.nc files)
- ability to read in different (newer) UFS IC format (related to using chgres_cube instead of global_chgres?)
- updated User's Guide to reflect changes
- include Shapely source code for installing on Hera since it isn't in the Anaconda2 module (any better way?)

This has been tested with Xia's ICs, example UFS IC data from the FTP site, other ICs from @llpcarson on Cheyenne, and the example ICs that ship with this repo.

Main changes in scm/etc/script/UFS_IC_generator.py and User's Guide chapters (ignore all the Shapely files for review).


llpcarson:
Grant - I think the name of the option/switch needs to be revised, this apparently is only indirectly related to the grib2 IC data source.  I took a quick look at the code that generates this, and an older version (global_chgres) has the "geolon" variable as an output field, while the current version does not have this field.  This newer version (chgres_cube) was selected *because* it can read grib2 ICs, but obviously there are other changes mixed in.  This should probably be the default - too, with an option to select the original behavior?

I'm not quite sure how to refer to this variant --- but grib2 may be confusing.

I should add that this is a quick-review hypothesis, which seems consistent with the test results, but may not be a complete description of what's happening!

grantfirl:
> Grant - I think the name of the option/switch needs to be revised, this apparently is only indirectly related to the grib2 IC data source. I took a quick look at the code that generates this, and an older version (global_chgres) has the "geolon" variable as an output field, while the current version does not have this field. This newer version (chgres_cube) was selected _because_ it can read grib2 ICs, but obviously there are other changes mixed in. This should probably be the default - too, with an option to select the original behavior?
> 
> I'm not quite sure how to refer to this variant --- but grib2 may be confusing.
> 
> I should add that this is a quick-review hypothesis, which seems consistent with the test results, but may not be a complete description of what's happening!

OK, I'll change this.

grantfirl:
OK, @llpcarson suggestion has been implemented and documented.

grantfirl:
@climbfuji @llpcarson @ligiabernardet When this is approved/merged, we're likely ready for beta02 and more testing.

grantfirl:
@climbfuji @llpcarson Before I merge, I have a question. I would rather not include Shapely source code in the SCM repository, but yesterday, I could not get Hera to install it using any of the normal ways that I knew of, hence the inclusion. However, I asked Xia and she (and subsequently I) have used the following:

`pip install --index-url http://anaconda.rdhpcs.noaa.gov/simple --trusted-host anaconda.rdhpcs.noaa.gov shapely --user`

So, I'd like to remove Shapely from the repo and use that. The question is, do we add this to the Hera setup scripts (after checking if it is installed first), or do we just paste this command in the User's Guide and tell folks to execute it if necessary? BTW, Cheyenne has shapely installed by default in the python environment that is used there.

climbfuji:
> @climbfuji @llpcarson Before I merge, I have a question. I would rather not include Shapely source code in the SCM repository, but yesterday, I could not get Hera to install it using any of the normal ways that I knew of, hence the inclusion. However, I asked Xia and she (and subsequently I) have used the following:
> 
> `pip install --index-url http://anaconda.rdhpcs.noaa.gov/simple --trusted-host anaconda.rdhpcs.noaa.gov shapely --user`
> 
> So, I'd like to remove Shapely from the repo and use that. The question is, do we add this to the Hera setup scripts (after checking if it is installed first), or do we just paste this command in the User's Guide and tell folks to execute it if necessary? BTW, Cheyenne has shapely installed by default in the python environment that is used there.

This is great - I'd say add it to the hera setup scripts, and provide instructions for users on generic platforms!

climbfuji:
@grantfirl maybe I missed that, but are there instructions for users (how) to install Shapely in the appropriate section supported platforms in the quick start chapter?

grantfirl:
> @grantfirl maybe I missed that, but are there instructions for users (how) to install Shapely in the appropriate section supported platforms in the quick start chapter?

Not all users will need to install shapely since it is only used in the UFS_IC_generator.py script. Therefore, I put the instructions in section 5.5. If you think that they should be moved, then we can do that.

climbfuji:
I think that's fine, sorry, I didn't see them (didn't look properly).

PR #171, opened 2020-03-12
Doc updates

ligiabernardet:
Mention that Noah-MP is not supported in introduction. Change UFS v1 release to UFS MR Weather App v1. Fix broken link.


grantfirl:
Thanks. I'm also just removing the stuff related to running with NoahMP in the UFS IC section.

PR #181, opened 2020-04-30
RRTMGP w/ GFS SDFs + some houskeeping

dustinswales:
This PR contains the necessary changes to work w/ ccpp-physics PR#446 (https://github.com/NCAR/ccpp-physics/pull/446).

Similar to that PR, there were some changes to the names of the radiation heating rates variables, so there is some cleaning up in GFS_typedefs addressing this.


scm/src/GFS_typedefs.F90
```diff
@@ -1850,12 +1850,6 @@ real (kind=kind_phys), pointer :: prsik (:,:) => null()   !< Exner function at i
     real (kind=kind_phys), pointer      :: cld_resnow(:,:)        => null()  !< Cloud snow effective radius
```
grantfirl, line 1850: @dustinswales I don't see these same changes in https://github.com/NCAR/fv3atm/pull/46. Were they previously committed there? We try to keep the two GFS_typedefs as close to each other as possible.

.gitmodules
```diff
@@ -4,5 +4,5 @@
 	branch = dtc/develop
```
grantfirl, line 4: After testing, we'd like to see this reverted.

scm/src/GFS_typedefs.F90
dustinswales: @grantfirl 
You are correct. I did some cleanup in here to the RRTMGP interstitial fields. I will do the same on the fv3atm side.

dustinswales:
SCM results, both SDFs, twpice configuration. RRTMGP (blue), RRTMG (red)
[scm_GvGP_v15p2.pdf](https://github.com/NCAR/gmtb-scm/files/4612166/scm_GvGP_v15p2.pdf)
[scm_GvGP_v16beta.pdf](https://github.com/NCAR/gmtb-scm/files/4612168/scm_GvGP_v16beta.pdf)

grantfirl:
@dustinswales Is this PR still valid? Since your RRTMGP changes went into master, I'm guessing that any PR to the SCM should also have started from its master? On your fork, I also see a scm_before_vacation branch. Does that one work with the ccpp-physics changes? I'd like to get whatever changes are required for the SCM to run the latest code that got merged into ccpp-physics/master for RRTMGP somehow, whether through this PR, or another, or just point me to some of your code so that I don't need to start from scratch.

dustinswales:
@grantfirl 
I need to revisit it this. I will cleanup and update this afternoon.

grantfirl:
> @grantfirl
> I need to revisit it this. I will cleanup and update this afternoon.

OK, let me know if you need anything. I'm happy to help, since this is required to have the master SCM code work with ccpp-physics/master.

dustinswales:
@grantfirl 
I updated to the NCAR master physics and added in some infrastructure to the SCM side, but I cannot get the prebuild to work.

grantfirl:
> @grantfirl
> I updated to the NCAR master physics and added in some infrastructure to the SCM side, but I cannot get the prebuild to work.

OK, I'll pull in your changes and see what I can do. If I find something that I can fix, I'll submit a PR from my fork to yours. I'm also just about ready to merge in #191, so I'll update your branch for that too. It shouldn't affect anything from your PR since it's just adding new cases.

grantfirl:
@dustinswales A quick update: starting from this branch, I now have a branch that is working OK. I'm building/compiling one last time to double-check, and then I'll submit a PR from my branch to this one for you to review. Where necessary, I took my lead from the associated changes in FV3 (like SDFs and ccpp_prebuild_config.py). You may want to pay particular attention to changes in those to make sure that I updated them as you intended.

grantfirl:
If you accept the PR from my fork to yours, I'd like to have you remove the Apollo setup script before I accept/merge. I think that the code managers want to keep only setup scripts for officially-supported machines.

dustinswales:
@grantfirl 
I accepted your PR. Builds/runs, everything works. I removed my local system configuration file.

ccpp/suites/suite_SCM_GFS_v15p2_ACM_ps.xml
```diff
@@ -33,7 +33,7 @@
       <scheme>GFS_suite_interstitial_2</scheme>
       <scheme>gmtb_scm_sfc_flux_spec</scheme>
       <scheme>GFS_PBL_generic_pre</scheme>
-      <scheme>bl_acm</scheme>
+      <!--<scheme>bl_acm</scheme>-->
```
climbfuji, line 36: Why is the ACM scheme commented out in the ACM suite?

scm/src/GFS_typedefs.F90
```diff
@@ -6968,6 +6984,17 @@ subroutine interstitial_phys_reset (Interstitial, Model)
       Interstitial%radar_reset = mod(Model%kdt-1, nint(Model%nsradar_reset/Model%dtp)) == 0
     end if
     !
+
+	!
+	! RRTMGP
+	!
+	if (Model%do_RRTMGP) then  
+	   if (Model%use_LW_jacobian) then 
+          !Interstitial%fluxlwUP_jac   = clear_val
```
climbfuji, line 6993: Why commented out? Same for line 6994.

ccpp/suites/suite_SCM_GFS_v15p2_ACM_ps.xml
grantfirl: This one is my fault. In a past PR to master, I included an ACM suite even though the ACM scheme is not in the ccpp-physics yet. I erroneously thought that it would "do no harm" since no one is forced to use it or anything, but I forgot that CMakeLists.txt is grabbing ALL suites in ccpp/suites to parse, which results in an error until the ACM scheme is in ccpp-physics. So, rather than remove the ACM suite altogether (since some folks in another project are actually using it with a non-master branch of ccpp-physics that has the ACM scheme), I decided to just comment this out for now so that ccpp_prebuild doesn't error out.

scm/src/GFS_typedefs.F90
grantfirl: @dustinswales I am curious about this too.

grantfirl:
I'll go ahead and merge. If @climbfuji would rather me remove the ACM SDF altogether, I can do that in a followup.

PR #185, opened 2020-07-10
Update master branch to use master branches of ccpp-physics and ccpp-framework

grantfirl:
This PR allows the SCM to work with the latest master branches of ccpp-physics and ccpp-framework. It also contains changes to work with HWRF physics, which will presumably be going to the master branches of the CCPP soon too.

Most of the changes are to facilitate new/different variables in the physics, although a few of the commits in July (specifically 75f3475) change the way multiple independent columns could be used. Rather than have multiple instances of GFS DDTs holding the different columns (and potentially allowing different physics suites to be run in each of them simultaneously), the horizontal dimension of the individual variables is used instead. E.g. instead of physics%Statein(i)%tgrs(1,k), you now have physics%Statein%tgrs(i,k). This closes the door on the potential to run multiple physics suites simultaneously, but this wasn't being done in practice due to module-level data anyway.

This PR also marks the ability to use the updated tendencies, although the non-physics tendency is not working due to memory management. This can be enabled in a future PR.

Lastly, the GFS_DCNV_generic_pre routines have been rearranged in many SDFs.


scm/src/GFS_typedefs.F90
```diff
@@ -113,6 +116,7 @@ module GFS_typedefs
     real(kind=kind_phys) :: dt_phys              !< physics  time step in seconds
 !--- restart information
     logical :: restart                           !< flag whether this is a coldstart (.false.) or a warmstart/restart (.true.)
+    logical :: cycling                           !< flag whether this is a coldstart (.false.) or a cycled run (.true.)
```
climbfuji, line 119: Note that this will be removed again in fv3atm in the near future, see https://github.com/NOAA-EMC/fv3atm/pull/140.

scm/src/GFS_typedefs.F90
```diff
@@ -3235,6 +3459,30 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
...
+    end if
+    if (Model%naux3d>0) then
+        allocate(Model%aux3d_time_avg(1:naux3d))
+        Model%aux3d_time_avg(1:naux3d) = aux3d_time_avg(1:naux3d)
+    end if
+    if (any(aux2d_time_avg) .or. any(aux3d_time_avg)) then
+      write(0,*) "Error, the SCM has not implemented time averaging of diagnostics in GFS_typedefs.F90"
+      stop
+    end if
+
```
climbfuji, line 3485: How is the time averaging done in SCM? The tendencies are accumulated in ccpp-physics. Are you subtracting the previous accumulated value after each time step?

scm/src/GFS_typedefs.F90
```diff
@@ -3290,6 +3537,15 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
     Model%fhlwr            = fhlwr
     Model%nsswr            = nint(fhswr/Model%dtp)
     Model%nslwr            = nint(fhlwr/Model%dtp)
+    if (restart) then
```
climbfuji, line 3540: Have you tried this in SCM? What differences do you see? Would be interesting to check (not for merging this PR, though).

scm/src/gmtb_scm.F90
```diff
@@ -92,66 +90,68 @@ subroutine gmtb_scm_main_sub()
     case default
       cdata_time_index = 2
   end select
+
```
climbfuji, line 93: Nice, this cleans up the code quite a bit.

scm/src/gmtb_scm_output.F90
```diff
@@ -219,6 +221,31 @@ subroutine output_init(scm_state)
   CALL CHECK(NF90_PUT_ATT(NCID=ncid,VARID=dummy_id,NAME="description",&
      VALUES="temperature tendency due to microphysics scheme"))
   CALL CHECK(NF90_PUT_ATT(NCID=ncid,VARID=dummy_id,NAME="units",VALUES="K s-1"))
+  CALL CHECK(NF90_DEF_VAR(NCID=ncid,NAME='dT_dt_ogwd',XTYPE=NF90_FLOAT,DIMIDS= (/ hor_dim_id, vert_dim_id, time_id /), &
```
climbfuji, line 224: It would be nice if we could suppress the output of the 3d tendencies if ldiag3d/qdiag3d are false.

scm/src/gmtb_scm_output.F90
```diff
@@ -318,8 +430,23 @@ subroutine output_init(scm_state)
   CALL CHECK(NF90_DEF_VAR(NCID=ncid,NAME='init_hour',XTYPE=NF90_FLOAT,VARID=hour_id))
   CALL CHECK(NF90_PUT_ATT(NCID=ncid,VARID=hour_id,NAME="description",VALUES="model initialization hour"))
   CALL CHECK(NF90_PUT_ATT(NCID=ncid,VARID=hour_id,NAME="units",VALUES=""))
-
-
+
```
climbfuji, line 433: We could do the same for the 3d tendencies, or?

scm/src/gmtb_scm_output.F90
```diff
@@ -426,187 +551,252 @@ subroutine output_append(scm_state, physics)
...
+  CALL CHECK(NF90_INQ_VARID(NCID=ncid,NAME="sw_rad_heating_rate",VARID=var_id))
+  CALL CHECK(NF90_PUT_VAR(NCID=ncid,VARID=var_id,VALUES=physics%Radtend%htrsw(:,:),START=(/1,1,scm_state%itt_out /)))
+  CALL CHECK(NF90_INQ_VARID(NCID=ncid,NAME="lw_rad_heating_rate",VARID=var_id))
+  CALL CHECK(NF90_PUT_VAR(NCID=ncid,VARID=var_id,VALUES=physics%Radtend%htrlw(:,:),START=(/1,1,scm_state%itt_out /)))
+  
+  CALL CHECK(NF90_INQ_VARID(NCID=ncid,NAME="dT_dt_lwrad",VARID=var_id))
   do i=1, scm_state%n_cols
-    dummy_1D(i) = physics%Interstitial(i)%dusfc1(1)
+    dummy_2D(i,:) = physics%Diag%dt3dt(i,:,1)/scm_state%dt
```
climbfuji, line 596: I see, this is where the time averaging happens for the 3d tendencies.

scm/src/gmtb_scm_output.F90
```diff
@@ -638,83 +828,104 @@ subroutine output_append(scm_state, physics)
...
   CALL CHECK(NF90_PUT_VAR(NCID=ncid,VARID=var_id,VALUES=dummy_1D,START=(/1,scm_state%itt_out /)))
   !
   do i=1, scm_state%n_cols
-     dummy_1D(i) = physics%Radtend(i)%sfcflw(1)%upfx0
+     dummy_1D(i) = physics%Radtend%sfcflw(i)%upfx0
   end do
   CALL CHECK(NF90_INQ_VARID(NCID=ncid,NAME="lw_dn_sfc_clr",VARID=var_id))
   CALL CHECK(NF90_PUT_VAR(NCID=ncid,VARID=var_id,VALUES=dummy_1D,START=(/1,scm_state%itt_out /)))
-
+
```
climbfuji, line 907: You could use the time_avg logical arrays in the same way as it is done in fv3atm - if it is .true. for a specific slice of the array, simply divide by the integration time (same as you do for the 3d tendencies).

scm/src/GFS_typedefs.F90
grantfirl: OK, thanks. No one would (should) ever change that from the default false value when running the SCM anyway.

scm/src/GFS_typedefs.F90
grantfirl: diag_phys_zero is called every time step and right before it is written out, these accumulated change variables are divided by the timestep to get the tendencies.

scm/src/GFS_typedefs.F90
grantfirl: Nope. Restarts are not implemented in the SCM since they are so short to run. This is another GFS_typedefs variable that is only there to minimize differences between FV3/SCM. There should probably be a list of namelist and other vars to not change from their default in the SCM.

scm/src/gmtb_scm_output.F90
grantfirl: What about if we just used a netCDF missing value if the logicals are false? Right now, it outputs zeros if they are false, but I suppose that could be misleading. I prefer to keep the variables actually in the output file so that scripts don't need to check for variables existing or not. I guess they would need to check for missing values instead though.

scm/src/gmtb_scm_output.F90
grantfirl: Sure. I could have an array of variable names and descriptions and shorten the code with a loop for sure. At some point, I was hoping to reimplement the output for the SCM though, so I'm not sure that it is worth it right now. I'd like to implement something where the users could specify which variables to write out by their CCPP standard names, implement time-averaging (instead of only instantaneous), etc.

scm/src/gmtb_scm_output.F90
grantfirl: OK, I'll look into that for the output rewrite.

grantfirl:
> Thanks for all the cleanup work, glad to see that the new ccpp-framework code is working as expected in SCM as well (when no blocking is used). I left several comments, but happy to approve if you think that it is not worth addressing them now.

Thanks for the comments and review. I'd like to go ahead and merge as-is and do a rewrite of the output routines later (which I had already planned on doing at some point anyway) taking your suggestions into account then. Mike Iacono is waiting for these changes and I need a good base to start some other development.

PR #187, opened 2020-07-17
Update for new RRTMG cloud overlap capability

mjiacono:
Changes to gmtb-scm:

- Source file GFS_typedefs.F90 was modified to add alpha to the type GFS_interstitial_type and to expand the description of iovr_lw and iovr_sw to include EXP and ER. 

This pull request is paired with a related pull request to NCAR/ccpp-physics #476.

### DIFF
--- a/ccpp/physics_namelists/input_GFS_v15p2.nml
+++ b/ccpp/physics_namelists/input_GFS_v15p2.nml
@@ -14,6 +14,8 @@
        iems           = 1
        iaer           = 111
        ico2           = 2
+       iovr_sw        = 5
+       iovr_lw        = 5
        isubc_sw       = 2
        isubc_lw       = 2
        isol           = 2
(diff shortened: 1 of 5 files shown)


ccpp/physics_namelists/input_GFS_v15p2.nml
```diff
@@ -14,6 +14,8 @@
        iems           = 1
        iaer           = 111
        ico2           = 2
+       iovr_sw        = 5
+       iovr_lw        = 5
```
ligiabernardet, line 18: I am not sure what is the default of iovr_sw and _lw, but we do not want to change the behavior of this suite. It should produce the same results as before

mjiacono:
The new options for iovr_lw/iovr_sw are 4 and 5.  I’ll reset those universally to 1 (MR overlap) in the revised updates.

Issue #186, opened 2020-07-13
Change in loading Anaconda module on Hera


grantfirl:
It has been reported that loading the Anaconda module on Hera has changed from:

```
module load contrib
module load anaconda/anaconda2
```

to 

```
module use -a /contrib/anaconda/modulefiles
module load anaconda/2.3.0
```

Verify and change as necessary.

grantfirl:
Addressed in #189

PR #189, opened 2020-07-24
Update SCM for masters of CCPP + Hera Anaconda fix

grantfirl:
- Update GFS_typedefs.[F90/meta] to match FV3 and to work with the latest masters of ccpp-physics and ccpp-framework
 - this allows compilation of FLAKE (although it has not been tested with the SCM -- I'm not sure it can be initialized as-is)
 - change module load statements for Anaconda python for Hera after machine change


grantfirl:
Address #186

scm/src/GFS_typedefs.F90
```diff
@@ -4104,8 +4110,13 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
        'max_lon=',max_lon,' max_lat=',max_lat,' min_lon=',min_lon,' min_lat=',min_lat,       &
        ' rhc_max=',Model%rhcmax
 
-!--- set nrcm
-    Model%nrcm = 2
+!--- set nrcm 
+
+    if (Model%ras) then
```
climbfuji, line 4115: This is not new, but I assume it was forgotten when the RAS updates were brought into SCM in the first place.

scm/src/GFS_typedefs.F90
```diff
@@ -1493,10 +1495,6 @@ real (kind=kind_phys), pointer :: prsik (:,:) => null()   !< Exner function at i
 
     !--- MP quantities for 3D diagnositics 
     real (kind=kind_phys), pointer :: refl_10cm(:,:) => null()  !< instantaneous refl_10cm
-
-    !--- Extra PBL diagnostics
```
climbfuji, line 1497: Is this a new change? Or just something we missed in an earlier update?

scm/src/GFS_typedefs.F90
grantfirl: Yes.

scm/src/GFS_typedefs.F90
grantfirl: I think it was missed. Just an extra variable that wasn't hurting anything.

PR #190, opened 2020-07-27
Tracer rework + Updates to work with PBL schemes needed in Lulin's DOE ASR project

grantfirl:
Major changes:
 - added externally-configured tracers (via files in scm/etc/tracer_config)

Notes on new tracer functionality:

 - tracer configuration files are txt files with each record having its name, CCPP standard name, and units (currently only the name is used)
 - Currently known tracers include the following names (all from FV3):
"sphum")
"o3mr")
"liq_wat")
"ice_wat")
"rainwat")
"snowwat")
"graupel")
"cld_amt")
"water_nc")
"ice_nc")
"rain_nc")
"snow_nc")
"graupel_nc")
"sgs_tke")
"liq_aero")
"ice_aero")
"q_rimef")
 - which tracers to use are specified at run-time: the run_gmtb_scm.py script now has the "-t" optional argument to specify the tracer configuration file; if not specified, the run script will look in default_tracers.py for a tracer configuration file corresponding to the specified suite. The run script also stages the tracer configuration file as "tracers.txt" in the run directory for the Fortran code to find.

Minor changes:

 - added SDFs and namelists to use operational GFS suite with YSU, saYSU, and ACM PBL schemes (prescribed surface only)
 - commented out ACM scheme in ccpp_prebuild_config.py so that this code can be used with current ccpp-physics master (ACM scheme to be added to ccpp-physics master later after modifying FV3)
 - added new index variables to GFS_typedefs.[F90/meta] corresponding to the array position of water vapor, cloud water, cloud ice, and ozone in the vertically-diffused tracer arrays (since they are different than their corresponding indices in the superset tracer array)


scm/etc/tracer_config/tracers_gsd.txt
```diff
@@ -0,0 +1,13 @@
...
+"liq_wat","cloud_condensed_water_mixing_ratio","kg kg-1"
+"ice_wat","ice_water_mixing_ratio","kg kg-1"
+"rainwat","rain_water_mixing_ratio","kg kg-1"
+"snowwat","snow_water_mixing_ratio","kg kg-1"
+"graupel","graupel_mixing_ratio","kg kg-1"
+"water_nc","cloud_droplet_number_concentration","kg-1"
+"ice_nc","ice_number_concentration","kg-1"
+"rain_nc","rain_number_concentration","kg-1"
+"o3mr","ozone_mixing_ratio","kg kg-1"
+"liq_aero","water_friendly_aerosol_number_concentration","kg-1"
```
climbfuji, line 11: How do you envisage using the Thompson non-aerosol aware version (`ltaerosol = .false.`)? A separate `tracers_gsd_no_aero.txt`?

grantfirl: > The PR itself is good to go, but I wish we had discussed this beforehand. If anything I would like the UFS to move away from specifying tracers (redundantly) in `field_table`. If we look at it, there is only one valid set of tracers that one can use for a chosen physics suite and some runtime flags (e.g. Thompson with/without aerosols), so specifying them redundantly in a `field_table` or other file isn't particulary useful, or? Sorry if I missed something, happy to stand corrected.

I'm not necessarily advocating for specifying them redundantly via an external file, and would certainly be open to these being controlled via a suite, e.g. via its SDF. This implementation was a response to some logic in GFS_typedefs that was dependent on the number of tracers. For the SCM, I had been using a superset of all tracers needed by all physics (that I have been tested with supported suites and some experimental suites). Automatically running with more tracers for some suites was causing the vertically-diffused tracer array to be unnecessarily large for some suites (and actively causing problems in some PBL schemes). By putting in this logic, I can run the SCM with "just enough" tracers. Once someone implements something different for tracers for FV3 (that is potentially CCPP suite-based), I'd love to switch to that. That definitely makes more sense to me too. It just doesn't exist today.

grantfirl: > How do you envisage using the Thompson non-aerosol aware version (`ltaerosol = .false.`)? A separate `tracers_gsd_no_aero.txt`?

Yep, a separate tracer config file.

climbfuji: > > The PR itself is good to go, but I wish we had discussed this beforehand. If anything I would like the UFS to move away from specifying tracers (redundantly) in `field_table`. If we look at it, there is only one valid set of tracers that one can use for a chosen physics suite and some runtime flags (e.g. Thompson with/without aerosols), so specifying them redundantly in a `field_table` or other file isn't particulary useful, or? Sorry if I missed something, happy to stand corrected.
> 
> I'm not necessarily advocating for specifying them redundantly via an external file, and would certainly be open to these being controlled via a suite, e.g. via its SDF. This implementation was a response to some logic in GFS_typedefs that was dependent on the number of tracers. For the SCM, I had been using a superset of all tracers needed by all physics (that I have been tested with supported suites and some experimental suites). Automatically running with more tracers for some suites was causing the vertically-diffused tracer array to be unnecessarily large for some suites (and actively causing problems in some PBL schemes). By putting in this logic, I can run the SCM with "just enough" tracers. Once someone implements something different for tracers for FV3 (that is potentially CCPP suite-based), I'd love to switch to that. That definitely makes more sense to me too. It just doesn't exist today.

Got it - in this case we should proceed with your PR, so that when the UFS gets cleaned up we do the same for the SCM.

grantfirl: Would it help to open an issue in fv3atm or ccpp-framework or somewhere else regarding migrating away from the field_table for tracers and implementing something in ccpp-framework to create a tracer array based on CCPP suites and other namelist options? I'm not sure 1) who would be responsible for implementing something and 2) which repo(s) it would need to touch.

ligiabernardet: My understanding is that NCAR (Steve Goldhaber) would be implementing the functionality of creating a tracer array based on CCPP suites in capgen (if the NCAR collaboration with NRL is finalized). In today's CCPP-Framework meeting, Steve said this is captured in an issue in the ccpp-framework repository, but I cannot find which issue.

climbfuji: https://github.com/NCAR/ccpp-framework/issues/291

PR #194, opened 2020-08-12
python3 compatibility for release branch

grantfirl:
1. add Dom's python 3 compatibility changes for run scripts from master branch
2. add python 3 compatibility for plot* scripts
3. change to python 3 for compute environments on Hera and Cheyenne

*edited

### DIFF
--- a/scm/etc/scripts/forcing_file_common.py
+++ b/scm/etc/scripts/forcing_file_common.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+
 from netCDF4 import Dataset
 import numpy as np
 
@@ -42,36 +44,36 @@ def ncdump(nc_fid, verb=True):
             a valid netCDF4.Dataset.variables key
         """
         try:
-            print "\t\ttype:", repr(nc_fid.variables[key].dtype)
+            print("\t\ttype:", repr(nc_fid.variables[key].dtype))
             for ncattr in nc_fid.variables[key].ncattrs():
-                print '\t\t%s:' % ncattr,\
-                      repr(nc_fid.variables[key].getncattr(ncattr))
+                print('\t\t%s:' % ncattr,\
+                      repr(nc_fid.variables[key].getncattr(ncattr)))
         except KeyError:
-            print "\t\tWARNING: %s does not contain variable attributes" % key
+            print("\t\tWARNING: %s does not contain variable attributes" % key)
 
     # NetCDF global attributes
     nc_attrs = nc_fid.ncattrs()
     if verb:
-        print "NetCDF Global Attributes:"
+        print("NetCDF Global Attributes:")
         for nc_attr in nc_attrs:
-            print '\t%s:' % nc_attr, repr(nc_fid.getncattr(nc_attr))
+            print('\t%s:' % nc_attr, repr(nc_fid.getncattr(nc_attr)))
     nc_dims = [dim for dim in nc_fid.dimensions]  # list of nc dimensions
     # Dimension shape information.
     if verb:
-        print "NetCDF dimension information:"
+        print("NetCDF dimension information:")
         for dim in nc_dims:
-            print "\tName:", dim
-            print "\t\tsize:", len(nc_fid.dimensions[dim])
+            print("\tName:", dim)
+            print("\t\tsize:", len(nc_fid.dimensions[dim]))
             print_ncattr(dim)
     # Variable information.
     nc_vars = [var for var in nc_fid.variables]  # list of nc variables
     if verb:
-        print "NetCDF variable information:"
+        print("NetCDF variable information:")
         for var in nc_vars:
             if var not in nc_dims:
-                print '\tName:', var
-                print "\t\tdimensions:", nc_fid.variables[var].dimensions
-                print "\t\tsize:", nc_fid.variables[var].size
+                print('\tName:', var)
+                print("\t\tdimensions:", nc_fid.variables[var].dimensions)
+                print("\t\tsize:", nc_fid.variables[var].size)
                 print_ncattr(var)
     return nc_attrs, nc_dims, nc_vars
 
--- a/scm/src/multi_run_gmtb_scm.py
+++ b/scm/src/multi_run_gmtb_scm.py
@@ -1,7 +1,13 @@
 #!/usr/bin/env python
 
+import sys
+PYTHON2 = sys.version_info[0] < 3
+
 import argparse
-import imp
+if PYTHON2:
+    from imp import load_source
+else:
+    from importlib.machinery import SourceFileLoader
 import os
 import logging
 import subprocess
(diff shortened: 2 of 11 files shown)


grantfirl:
This has been tested to work using the updated machine setup scripts on Cheyenne and Hera (both .csh and .sh versions). The plotting scripts run though fine using all python 3 now.

scm/etc/scripts/forcing_file_common.py
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
```
climbfuji, line 3: this is of course only required for Python2, but it won't hurt with Python3.

scm/src/multi_run_gmtb_scm.py
```diff
@@ -1,7 +1,13 @@
 #!/usr/bin/env python
 
+import sys
+PYTHON2 = sys.version_info[0] < 3
```
climbfuji, line 4: Argh. This is what I did for `ccpp_prebuild.py` and it works with both Python 2 and 3. Not sure if it would work in your case:
```
import importlib
...
    # Import the host-model specific CCPP prebuild config;
    # split into path and module name for import
    configpath = os.path.abspath(os.path.dirname(configfile))
    configmodule = os.path.splitext(os.path.basename(configfile))[0]
    sys.path.append(configpath)
    ccpp_prebuild_config = importlib.import_module(configmodule)
```

scm/src/run_gmtb_scm.py
```diff
@@ -83,15 +83,15 @@ def execute(cmd):
     status = p.returncode
     if status == 0:
         message = 'Execution of "{0}" returned with exit code {1}\n'.format(cmd, status)
-        message += '    stdout: "{0}"\n'.format(stdout.rstrip('\n'))
-        message += '    stderr: "{0}"'.format(stderr.rstrip('\n'))
+        message += '    stdout: "{0}"\n'.format(stdout.rstrip('\n'.encode()))
```
climbfuji, line 86: I thought you need to use `decode`, but maybe I am wrong.

I ended up using the following code (see ccpp-framework's `common.py` script):
```
def execute(cmd, abort = True):
    """Runs a local command in a shell. Waits for completion and
    returns status, stdout and stderr. If abort = True, abort in
    case an error occurs during the execution of the command."""

    # Set debug to true if logging level is debug
    debug = logging.getLogger().getEffectiveLevel() == logging.DEBUG

    logging.debug('Executing "{0}"'.format(cmd))
    p = subprocess.Popen(cmd, stdout = subprocess.PIPE,
                         stderr = subprocess.PIPE, shell = True)
    (stdout, stderr) = p.communicate()
    status = p.returncode
    if debug:
        message = 'Execution of "{0}" returned with exit code {1}\n'.format(cmd, status)
        message += '    stdout: "{0}"\n'.format(stdout.decode(encoding='ascii', errors='ignore').rstrip('\n'))
        message += '    stderr: "{0}"'.format(stderr.decode(encoding='ascii', errors='ignore').rstrip('\n'))
        logging.debug(message)
    if not status == 0:
        message = 'Execution of command {0} failed, exit code {1}\n'.format(cmd, status)
        message += '    stdout: "{0}"\n'.format(stdout.decode(encoding='ascii', errors='ignore').rstrip('\n'))
        message += '    stderr: "{0}"'.format(stderr.decode(encoding='ascii', errors='ignore').rstrip('\n'))
        if abort:
            raise Exception(message)
        else:
            logging.error(message)
    return (status, stdout.decode(encoding='ascii', errors='ignore').rstrip('\n'),
                    stderr.decode(encoding='ascii', errors='ignore').rstrip('\n'))
```

scm/src/multi_run_gmtb_scm.py
grantfirl: Hmm. This was just brought over from your original python 3 implementation of the run scripts. You're saying that I should switch to this? I tried this in my python 2 environment and I'm getting an ImportError exception, with or without appending the path. In practice, I had been allowing a relative path to a filename as input into this script, and if the filename starts with ../, I'm also getting an error about needing to add a package to the import_module() call.

scm/src/run_gmtb_scm.py
grantfirl: Again, this was copied from the master branch that you originally did. This seemed to work on every machine with python2/3 that I tried.

scm/etc/scripts/forcing_file_common.py
grantfirl: Yep, I thought that we were trying to make any python scripts compatible with 2 AND 3.

scm/src/multi_run_gmtb_scm.py
grantfirl: It seems like there a bunch of ways to do this, but none seem superior to others to me and work for python 2 and 3: https://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path

scm/src/run_gmtb_scm.py
climbfuji: Correct, but I had to make changes afterwards because it failed on stampede (and I think also one of the wcoss platforms). Can you test on stampede?

climbfuji: Also, the error would only show up if that particular piece of code is executed (since Python is completely ignoring code that doesn't get executed).

scm/src/multi_run_gmtb_scm.py
climbfuji: I just thought that
```
import importlib
...
importlib.import_module(mymodule)
```
would be a cleaner way, but whatever works is ok.

scm/src/run_gmtb_scm.py
grantfirl: No, I don't have access to stampede. I can proactively change to this and retest though.

climbfuji: I can check that quickly for you, hang on.

scm/src/multi_run_gmtb_scm.py
grantfirl: The following code works on my Mac python 2/3 versions (copied from an answer on StackExchange). I'm not sure why it wasn't working with your suggested code since it is very similar.

import importlib

dirname, basename = os.path.split(args.file) 
sys.path.append(dirname) 
module_name = os.path.splitext(basename)[0] 
scm_runs = importlib.import_module(module_name) 
sys.path.pop()

scm/src/run_gmtb_scm.py
climbfuji: I could not reproduce the error we saw with the UFS on stampede, so I suggest to leave it as it is.

grantfirl: It was easy to change, and I retested with python 2 and 3 with success, so I committed this as you suggested.

PR #198, opened 2020-08-31
update SCM to work with ccpp-physics master (8/31/2020)

grantfirl:
update GFS_typedefs.[F90/meta] and ccpp/physics_namelists (to remove do_sfcperts) to work with latest ccpp-physics master


grantfirl:
@dustinswales After this merge, the master should be up-to-date with ccpp-physics.

PR #203, opened 2020-09-23
remove resetting NETCDF environment variable after loading NCEPLIBS; …

grantfirl:
…add special NetCDF treatment for running on Hera.intel in CMakeLists.txt


climbfuji (approved):
I see you fixed it!

grantfirl:
> I see you fixed it!

It seems to work, although I'm not sure it's the "correct" way to do things. I consider myself a butcher when it comes to build stuff. For example, specifying netCDF on Hera is different than every other platform (using the FIND cmake module rather than just pointing to an installation with an environment variable), which seems a bit weird to me. Also, should I make CMakeModules a submodule under gmtb-scm to have the latest FindNetCDF.cmake at all times?

climbfuji:
> > I see you fixed it!
> 
> It seems to work, although I'm not sure it's the "correct" way to do things. I consider myself a butcher when it comes to build stuff. For example, specifying netCDF on Hera is different than every other platform (using the FIND cmake module rather than just pointing to an installation with an environment variable), which seems a bit weird to me. Also, should I make CMakeModules a submodule under gmtb-scm to have the latest FindNetCDF.cmake at all times?

I think this is more than good enough for the public release. We can improve the build system as we go forward with master.

PR #207, opened 2020-09-28
Add TKE as a tracer for GFS_v16beta-based suites

grantfirl:
A problem was reported by Qingfu Liu and verified by @dustinswales where the ozone profile contained nonsensical values at some point early in the simulation, after the initialization step. This issue was tracked to the tracer index for TKE being set to the same value as the tracer index for ozone in some physics configurations. In this case, a suite with a PBL scheme that expects the TKE tracer to be present was being called without a TKE tracer being defined. This ultimately points to an issue in the setting of tracer indices in GFS_typedefs, which would benefit from more complete consistency checks between specified physics schemes and provided tracers, or a simplified reimplementation of tracer indices in general.

As a quick fix, TKE was added as a tracer for suites in the SCM that use the satmedmf(q) scheme -- those based on the GFS_v16beta suite.

This was tested to fix the issue on Hera and Mac.


climbfuji (approved):
Approved. But I find it nuts that tracers are defined in an external file `field_table` in FV3 (and mirrored here in `tracers_*.txt`, because we want to be consistent) when the choice of physics defines clearly which tracers are required!

grantfirl:
100% agreed. I'm looking forward to this problem being solved "for real" at some point, perhaps after the capgen.py transition.

PR #214, opened 2020-11-11
Update SCM suites to match FV3

grantfirl:
This PR builds on https://github.com/NCAR/gmtb-scm/pull/213 (should be merged afterward).

It only updates SCM suites to function like FV3-based ones. No changes in SDFs were detected, but several changes in namelist settings were. They include:

- add qdiag3d = T for all suites (SCM-only change, since we always want to output moisture tendencies)
- add effr_in = T for the GSD_v1 suite
- set iccn = 1 in CSAWMG suite


climbfuji (approved):

grantfirl:
addresses https://github.com/NCAR/gmtb-scm/issues/212

PR #215, opened 2020-11-11
Move v4.1 release updates to master

grantfirl:
This PR should follow #213 and #214 (it contains those changes as well). Its contents have been reviewed before as part of the v4.1 release. This PR is merely merging those changes (primarily python3 and build stuff) back into master.

Update on 2020/11/13:
The namelist variables iovr_lw and iovr_sw were changed to iovr in the GFS_v16beta and HWRF suites. This was supposed to happen with #213, but I missed this.

PR #213, opened 2020-11-10
Update for SCM to use HWRF physics

grantfirl:
This update does the following:

- adds HWRF physics files to ccpp_prebuild_config.py
- adds HAFS_v0_hwrf and HAFS_v0_hwrf_thompson SDFs, tracer configurations, and physics input namelists, following UFS RTs using these suites (and adds to default_namelists.py and default_tracers.py)
- adds surface tables for HWRF surface physics
- updates GFS_typedefs.[F90,meta] to match FV3 changes for HWRF physics

PR #216, opened 2020-11-12
Add RRFS suite

grantfirl:
This PR builds upon #213, #214, and #215 and simply adds the RRFS_v1beta suite.

Method:
- copy FV3_RRFS_v1beta SDF from FV3 repo, develop branch
- start with namelist from NOAA-EMC/regional_workflow repository, develop branch (found in regional_workflow/ush/templates/input.nml.FV3)
- remove all namelists in that file except for gfs_physics_nml and cires_ugwp_nml
- modified namelist according to changes for RRFS_v1beta found in regional_workflow/ush/templates/FV3.input.yml
  - changes include lsm=2, lsoil_lsm=4, imfshalcnv=-1, imfdeepcnv=-1, do_mynnsfclay=.true., do_deep = .false.
  - note that the namelist above contains a couple of errors: do_sfcperts is no longer a valid namelist variable and effr_in is not set to true, despite it using Thompson; both of these "errors" are addressed in the SCM version
 - the tracer configuration is constructed from the field_table found in regional_workflow/ush/templates/field_table.FV3_RRFS_v1beta

Both the standard and "ps" (prescribed surface) versions of the SDF are added. The default namelists and tracers are updated to use the new files with the RRFS_v1beta suite as well. 

This suite was tested with all currently supported cases, and it compiled/ran. No effort was done to check for scientific "correctness" yet. In fact, using the SCM for deep convective cases without a convective parameterization is not necessarily expected to give good results. This suite may be limited to non-convective SCM cases or to use with UFS initial conditions and forcing when that is available, as any resolved convective activity should show up as an advective forcing term when FV3 is run at a fine enough horizontal grid-spacing.


grantfirl:
@climbfuji @llpcarson This PR is ready for review.

climbfuji (approved):

Issue #210, opened 2020-10-26
Setting surface temperature from forcing file even with UFS ICs


grantfirl:
Qingfu Liu pointed out that the surface temperature is being set from the forcing data. While this is appropriate for observationally-based cases that use specified surface fluxes (and even perhaps those that do not), it is not appropriate for when the SCM is using UFS ICs -- the surface temperature should be allowed to vary according to the surface energy budget as calculated in the physics.

Issue #209, opened 2020-10-21
Transient in surface temperature when using LSM (even with UFS ICs)


grantfirl:
Qingfu Liu discovered that the surface temperature goes down to 150C shortly after initialization, even when using UFS initial conditions to initialize the Noah LSM. This resembles a strong spinup-related transient, although I would expect there not to be LSM spinup when using UFS ICs. Check whether LSM variables are receiving the correct data at initialization time.

grantfirl:
weizhong.zheng@noaa.gov is also interested in a solution.

PR #217, opened 2020-12-03
Update initialization of surface related variables

grantfirl:
This PR accomplishes the following:
 - Addresses issues #209 and #210. 
 - Also addresses bug report by @wangevine related to specified surface fluxes being zero as long as used with new ccpp-physics PR that fixes gmtb_scm_sfc_flux_spec scheme.
 - Updates SCM to use latest master branches of ccpp-physics/framework

This represents a major rewrite of surface variable initialization to address the above issues and make future maintenance easier.

UFS_IC_generator.py
 - reads in ALL variables from surface file (as opposed to only some) following what is done in FV3GFS_io.F90; if variables are not present in the file, it will still write out the variable with values equal to a user-configurable missing value
 - handles all Noah LSM, RUC LSM, NoahMP LSM, and NSST variables
 - handles new fractional landmask and lake variables in the 'oro' files
 - outputs all variables (including missing ones)

gmtb_scm_input.F90
 - reads in all new variables output by UFS_IC_generator.py
 - utilizes new NetCDF_read_var module to handle missing data

gmtb_scm_type_defs.F90
 - removes redundant variables in the 'state' DDT introduced when UFS ICs were originally added
 - adds new variables to 'input' DDT only
 - new "set" type-bound procedure for the physics DDT (should be used only for initialization of physics variables that doesn't happen in the ccpp_physics_init stage; should NOT be used for variables that are associated with pointers that can be modified/forced outside of the physics by the SCM)
 - updated "associate" type-bound procedure of physics DDT to only have pointer association for variables that get modified/forced by SCM (specifically surface temperature logic)
 - new "set" routine conditionally sets Sfcprop variables, depending on missing data (will STOP if variables are required but missing)
 - includes logic from FV3/io/FV3GFS_io.F90 to deriving surface variables (related to fractional landmask, lake, NSST, NoahMP, and RUC LSM)
 - includes NoahMP "cold start" logic when NoahMP is active and Noah LSM input data is present, but NoahMP input data is missing

gmtb_scm_utils.F90
 - new NetCDF read module with generic interfaces for reading variables or setting to missing
 - new module for quality controlling variables or stopping if encountering missing values for required variables

gmtb_scm.F90
 - includes call to new physics%set routine after the physics%associate call


grantfirl:
Testing was done for:
 - existing specified surface flux cases (fixes issue noticed by @wangevine)
 - existing cases with simple ocean
 - existing UFS IC cases (also tested with IC files that have fractional landmask and lake data, but not RUC data yet)

Note to GSL folks @climbfuji @tanyasmirnova This PR should pave the way for testing with the RUC LSM. If there are FV3 IC files somewhere (after processing by global_chgres) that have RUC LSM variables, I could test it.

climbfuji:
RUC can be / is usually initialized from ICs that were created with Noah LSM - but both should work.

scm/src/gmtb_scm.F90
```diff
@@ -37,11 +37,11 @@ subroutine gmtb_scm_main_sub()
   character(len=16) :: logfile_name
 
   call get_config_nml(scm_state)
-
+
```
climbfuji, line 40: These extra whitespaces are not required, better to remove (in total four places in this file).

scm/src/CMakeLists.txt
```diff
@@ -299,12 +299,16 @@ ADD_DEPENDENCIES(ccppphys ccpp)
 
 SET(scm_source_files gmtb_scm.F90
             gmtb_scm_input.F90
-            gmtb_scm_utils.F90
+            #gmtb_scm_utils.F90               #gets compiled with CCPP
```
climbfuji, line 302: I was confused at first by "gets compiled with CCPP", of course it means that this is a dependency in CCPP. I would just remove those few lines altogether.

climbfuji (commented):
I don't claim to understand or to have reviewed line by line the surface initialization, but at least I didn't see anything obviously wrong. Some of the new netCDF routines could have been collapsed into fewer by using elemental functions/subroutines, but the current implementation is fine.

climbfuji (approved):

scm/src/CMakeLists.txt
grantfirl: addressed in last commit

scm/src/gmtb_scm.F90
grantfirl: addressed in last commit

PR #222, opened 2021-01-21
DEPHY input format (plus forcing rework)

grantfirl:
This pull request represents the initial implementation of version 1 of the DEPHY-SCM input format (https://docs.google.com/document/d/118xP04jB9HO7Y2LqWk3HZpZ9n3CFujgzimLI7Ug8vO4/edit?pli=1
). This does not remove support for the existing input format. In a future PR, I will provide a script to convert cases in the existing format to the DEPHY format and include existing cases in the DEPHY format. At that time, it will be possible to remove support for the older format.

This PR was tested with all of the cases in the DEPHY GitHub repo (master branch) found here: 
https://github.com/GdR-DEPHY/DEPHY-SCM
In order for others to test the format, please clone the above repository under the scm/data directory. At some point, it may make sense to make this another submodule of the top level CCPP SCM, but it is not as of this PR. When the DEPHY-SCM repo is cloned, one can make use of the following new cases, with case configuration files appearing in scm/etc/case_config:

1. AMMA_REF
2. ARMCU_E3SM, ARMCU_MESONH, ARMCU_REF
3. AYOTTE_00SC (does not work due to inconsistency in case file - see https://github.com/GdR-DEPHY/DEPHY-SCM/issues/1)
4. BOMEX_REF (does not work due to inconsistency in case file - see https://github.com/GdR-DEPHY/DEPHY-SCM/issues/1)
5. DYNAMO_NSA3a (2-month-long case!), DYNAMO_NSA3a_D1
6. GABLS1_REF (does not work due to inconsistency in case file - see https://github.com/GdR-DEPHY/DEPHY-SCM/issues/1)
7. IHOP_REF
8. MPACE_REF
9. RICO_MESONH
10. SANDU_FAST,SLOW,REF
11. SCMS_REF

Except where noted, all cases were run successfully to completion and some attempt was made to look at plots for the "reasonableness" test.

As part of the DEPHY-format case files, there are global NetCDF attributes that control how the forcing is applied. Some of the global attributes in the DEPHY-format case files effectively replaces namelist variables that are part of the existing case setup in scm/etc/case_config. The namelist variables that are superseded include the following:

1. runtime (deduced from end_date - start_date in the DEPHY-format files)
2. thermo_forcing_type (replaced by individual control of forcing terms)
3. mom_forcing_type (replaced by individual control of forcing terms)
4. relax_time (replaced by individual control of forcing terms)
5. sfc_flux_spec (replaced by individual control of forcing terms)
6. sfc_roughness_length (replaced by global attribute)
7. sfc_type (replaced by global attribute)
8. year, month, day, hour, min (replaced by state date global attribute)

Most of the coding changes to support the new format happen in gmtb_scm_input.f90 and gmtb_scm_forcing.f90.

gmtb_scm_input.f90 changes:

- add reference_profile_dir (used to assume the same as the processed_case_dir, but this causes an issue when DEPHY-SCM is a submodule)
- read NetCDF files that follow the DEPHY-SCM v1 format
- there is no attempt to read in LSM ICs from DEPHY-format files, since LSM-related variables are not in the standard; this will need to be an extension of the format
- conditionally reads variables based on global attributes in the file
- added logical checks for input data based on global attributes
- although the format is capable of support >1 column, it is assumed that the first column will be used for now (does not affect existing DEPHY-SCM cases)
- convert mixing ratios to specific humidities as necessary
- added logic to initialize all of qt, qv, ql, and qi, depending on what is provided in the file
- handle the specification of absolute temperature, potential temperature, and liquid-water potential temperature (since all are supported and any/all may appear in DEPHY-format files)
- handle prescribed surface fluxes in both kinematic and W m-2
- handle separate radiation forcing or radiation forcing included in the temperature advection term (new "no_rad" flag needed in ccpp-physics)
- handle nudging to profiles above a given pressure or height

gmtb_scm_forcing.f90 changes:

- changes to interpolate_forcing() subroutine
 -- keep existing code for old format, but add code for DEPHY-format
 -- only interpolate in space/time based on following new control variables:
  --- force_omega or force_w (force_omega has higher precedent)
  --- forge_geo (geostrophic wind forcing)
  --- force_adv_T (total advective term for temperature, potential temperature, or liquid potential temperature)
  --- force_adv_qv (total advection term for moisture)
  --- force_adv_u,v (total advection term for u,v)
  --- force_nudging_T (uses either temperature or liquid potential temperature profiles)
  --- force_nudging_qv
  --- force_nudging_u,v
  --- surface_thermo_control (0=> prescribed surface fluxes in kinematic form 1=> prescribed surface fluxes in W m-2 2=> prescribed SST for simple ocean model, 3=> LSM)
  -- surface_momentum_control (0=> given roughness length, 1=> time-varying friction velocity (implementation TODO)
- new apply_forcing_DEPHY() subroutine
 -- forcing is modular (all forcing terms are additive into a total forcing tendency term)
 -- force_omega is used if possible, otherwise force_w if present
  --- can use "use_theta" switch to convert to potential temperature before calculating vertical advection term (avoids having to calculate a separate adiabatic compression/expansion term)
  --- if not using "use_theta" the adiabatic compression/expansion term is calculated for straight vertical temperature advection
  -- nudging is applied only above the specified level

other changes:

- main gmtb_scm.f90 file
 -- added call to new case_init subroutine
 -- set the new no_rad variable if necessary
 -- call new apply_forcing routine
- gmtb_scm_setup.F90
 -- handle T and liquid potential temperature ICs as well as ql and qi
- gmtb_scm_time_integration.f90 file
 -- call new apply_forcing subroutine
- gmtb_scm_type_defs.f90 file
 -- allocate/initialize new control and forcing variables
- gmtb_scm_utils.f90 file
 -- extend NetCDF_read_var to single/double precision and more ranks
 -- add conditionally_read_var subroutine try to read variables from NetCDF and return missing if not present
- run_gmtb_scm.py run script
 -- checks the DEPHY-format input file for use of prescribed surface fluxes to properly substitute prescribed-surface flux SDF

TODO in followup PRs:
- DEPHY-SCM does not have initial profiles of trace gases for radiation, particularly ozone; need to implement a separate solution
- implement a time-varying friction velocity in addition to specified roughness length

### DIFF
--- a/scm/src/gmtb_scm.F90
+++ b/scm/src/gmtb_scm.F90
@@ -49,9 +49,8 @@ subroutine gmtb_scm_main_sub()
       stop
   end select
   
-
   call get_reference_profile(scm_state, scm_reference)
-
+  
   select case(trim(adjustl(scm_state%model_name)))
     case("GFS")
       !>  - Call get_GFS_grid in \ref vgrid to read in the necessary coefficients and calculate the pressure-related variables on the grid.
@@ -141,8 +140,10 @@ subroutine gmtb_scm_main_sub()
   
   !override radiation frequency
   if (scm_state%force_rad_T > 0) then
-    !set a flag that doesn't allow lsswr/lslwr to be set to true in GFS_time_vary_pre (that the CCPP knows about)
-    physics%Model%no_rad = .true.
+    !turn off radiation since it is already accounted for in the forcing
+    !set radiation calling periods to -1 in order to prevent lsswr/lslwr from being be set to true in GFS_time_vary_pre
+    physics%Model%nsswr = -1
+    physics%Model%nslwr = -1
   end if
   
   !override fhzero in physics namelist if n_itt_diag is set in the case namelist
@@ -323,6 +324,7 @@ subroutine gmtb_scm_main_sub()
   !end if
 
   do i = 2, scm_state%n_timesteps
+    
     !are we still in spinup mode?
     if (scm_state%do_spinup .and. i <= scm_state%spinup_timesteps) then
       in_spinup = .true.
@@ -349,18 +351,18 @@ subroutine gmtb_scm_main_sub()
       scm_state%temp_u = scm_state%state_u
       scm_state%temp_v = scm_state%state_v
     end if
-
+    
     call interpolate_forcing(scm_input, scm_state, in_spinup)
     
     call calc_pres_exner_geopotential(1, scm_state)
-
+    
     !zero out diagnostics output on EVERY time step - breaks diagnostics averaged over many timesteps
     !call physics%Diag%rad_zero(physics%Model)
     !call physics%Diag%phys_zero(physics%Model)
 
     !pass in state variables to be modified by forcing and physics
     call do_time_step(scm_state, physics, cdata, in_spinup)
-
+    
     if (scm_state%time_scheme == 2) then
       !for filtered-leapfrog scheme, call the filtering routine to calculate values of the state variables to save in slot 1 using slot 2 vars (updated, unfiltered) output from the physics
       call filter(scm_state)
@@ -387,6 +389,7 @@ subroutine gmtb_scm_main_sub()
       end if
 
     end if
+    
   end do
 
   call ccpp_physics_finalize(cdata, suite_name=trim(trim(adjustl(scm_state%physics_suite_name))), ierr=ierr)
--- a/scm/src/gmtb_scm.F90
+++ b/scm/src/gmtb_scm.F90
@@ -38,8 +38,17 @@ subroutine gmtb_scm_main_sub()
   logical                           :: in_spinup
 
   call get_config_nml(scm_state)
-
-  call get_case_init(scm_state, scm_input)
+  
+  select case(scm_state%input_type)
+    case(0)
+      call get_case_init(scm_state, scm_input)
+    case(1)
+      call get_case_init_DEPHY(scm_state, scm_input)
+    case default
+      write(*,*) 'An unrecognized specification of the input_type namelist variable is being used. Exiting...'
+      stop
+  end select
+  
 
   call get_reference_profile(scm_state, scm_reference)
 
@@ -64,7 +73,7 @@ subroutine gmtb_scm_main_sub()
   end select
 
   !allocate(cdata_cols(scm_state%n_cols))
-
+  
   call set_state(scm_input, scm_reference, scm_state)
 
   call calc_geopotential(1, scm_state)
@@ -129,6 +138,12 @@ subroutine gmtb_scm_main_sub()
                        physics%Diag, physics%Interstitial, 0, 1, 1,                &
                        physics%Init_parm, scm_state%n_cols, scm_state%lon,         &
                        scm_state%lat, scm_state%area)
+  
+  !override radiation frequency
+  if (scm_state%force_rad_T > 0) then
+    !set a flag that doesn't allow lsswr/lslwr to be set to true in GFS_time_vary_pre (that the CCPP knows about)
+    physics%Model%no_rad = .true.
+  end if
   
   !override fhzero in physics namelist if n_itt_diag is set in the case namelist
   if (scm_state%n_itt_diag >= 1) then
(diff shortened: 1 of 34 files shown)


grantfirl:
@wangevine This is the pull request that I referenced in the forum. It is not finished, but it is getting close.

grantfirl:
There is no longer a need to have a companion ccpp-physics PR, since the "no_rad" variable is no longer needed to turn off the radiation.

scm/src/gmtb_scm.F90
```diff
@@ -38,11 +38,19 @@ subroutine gmtb_scm_main_sub()
...
+    case(1)
+      call get_case_init_DEPHY(scm_state, scm_input)
+    case default
+      write(*,*) 'An unrecognized specification of the input_type namelist variable is being used. Exiting...'
+      stop
+  end select
+  
   call get_reference_profile(scm_state, scm_reference)
-
+
```
climbfuji, line 53: unnecessary whitespace

scm/src/gmtb_scm.F90
```diff
@@ -38,11 +38,19 @@ subroutine gmtb_scm_main_sub()
   logical                           :: in_spinup
 
   call get_config_nml(scm_state)
-
-  call get_case_init(scm_state, scm_input)
-
+
```
climbfuji, line 41: I wonder if, at some point, we should switch to strings instead of cryptic integers.

scm/src/gmtb_scm.F90
```diff
@@ -64,7 +72,7 @@ subroutine gmtb_scm_main_sub()
   end select
 
   !allocate(cdata_cols(scm_state%n_cols))
-
+
```
climbfuji, line 75: ditto

scm/src/gmtb_scm.F90
```diff
@@ -130,6 +138,14 @@ subroutine gmtb_scm_main_sub()
                        physics%Init_parm, scm_state%n_cols, scm_state%lon,         &
                        scm_state%lat, scm_state%area)
   
+  !override radiation frequency
+  if (scm_state%force_rad_T > 0) then
+    !turn off radiation since it is already accounted for in the forcing
+    !set radiation calling periods to -1 in order to prevent lsswr/lslwr from being be set to true in GFS_time_vary_pre
+    physics%Model%nsswr = -1
```
climbfuji, line 145: glad to see that this works

scm/src/gmtb_scm.F90
```diff
@@ -330,18 +351,18 @@ subroutine gmtb_scm_main_sub()
       scm_state%temp_u = scm_state%state_u
       scm_state%temp_v = scm_state%state_v
     end if
-
+
```
climbfuji, line 354: unnecessary whitespaces (several more in this file)

scm/src/gmtb_scm_utils.F90
```diff
@@ -78,6 +78,24 @@ subroutine interpolate_to_grid_centers(n_input_levels, input_pres, input_field,
   !> @}
 end subroutine
 
+subroutine find_state_vertical_index_from_input(input_pres, pres_l, k_in, k_out)
```
climbfuji, line 81: Wonder if this routine and `find_vertical_index_pressure` from `gmtb_scm_input.F90` could be consolidated into one?

climbfuji (approved):
This is an impressive amount of work and a lot to review. I did not check every single line obviously, but this looks good to me. Just a few questions and suggestions for future improvements.

scm/src/gmtb_scm.F90
grantfirl: Sure, one could do that. I think that I subconsciously avoid using strings as much as possible in Fortran due to them being difficult to handle in the past. In this particular case, though, I'm hoping to drop the need for the input_type altogether eventually when ONLY the DEPHY format is supported.

scm/src/gmtb_scm_utils.F90
grantfirl: Yes, I think that you're right. The find_vertical_index_pressure was supposed to find the vertical index of a given pressure (or first below a given pressure). This one is finding the vertical index in one vertical array of pressure given a particular index in another vertical array of pressure, but it would be identical to just use the first one and give it input_pres(k_in) instead of the whole input_pres array and an index to use.

grantfirl: @climbfuji See https://github.com/NCAR/gmtb-scm/pull/222/commits/f30731363b1d8bacf0ad673664c68f7b46edd48d for consolidated solution.

scm/src/gmtb_scm.F90
grantfirl: Yes, thanks for the suggestion -- always nice to avoid new variables if possible.

climbfuji (approved):

PR #223, opened 2021-01-27
implement timestep_init phases and catch up with master of ccpp-physics

grantfirl:
- Implement the timestep_init/finalize phases in SCM source code
- update to ccpp-physics master as of 2021/01/26
- change build system so that ccpp_prebuild.py writes caps/cmake snippets to build directory rather than src directory (this now works with the default CMakeLists.txt in ccpp-physics)
- change the diagnostic bucket clearing to SCM code
-- change timing of diagnostic clearing (order of operations was wrong)
-- allow for diagnostic clearing every timestep
- comment out reading in RUC LSM init variables since changes to ccpp-physics has rendered it obsolete (followup PR needed to read in RUC LSM init data)


grantfirl:
Associated PR: https://github.com/NCAR/ccpp-physics/pull/562

grantfirl:
The fix in the CMakeLists.txt for ccpp-physics in https://github.com/NCAR/ccpp-physics/commit/2d8c061a3f6aead4e8e8184095fad5a049009429#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a leads to build issues in the SCM, but it is likely the SCM build system's fault.

grantfirl:
Also, the RUC LSM code that was commented out needs to be updated following a recent change in that code.

scm/src/GFS_typedefs.meta
```diff
@@ -607,6 +614,14 @@
   dimensions = (horizontal_loop_extent)
   type = real
   kind = kind_phys
+[sncovr_ice]
+  standard_name = surface_snow_area_fraction_over_ice
+  long_name = surface snow area fraction over ice
+  units = frac
+  dimensions = (horizontal_dimension)
```
climbfuji, line 621: Why is this `horizontal_dimension`? Should be `horizontal_loop_extent` I think. Same for a few other variables in this `.meta` file (search & replace?).

scm/src/gmtb_scm_time_integration.F90
```diff
@@ -113,13 +117,42 @@ subroutine do_time_step(scm_state, physics, cdata, in_spinup)
       endif
     endif
   end do
-
+
```
climbfuji, line 120: Seeing that the same code exists twice, once in `scm/src/gmtb_scm_time_integration.F90` and once in `scm/src/gmtb_scm.F90`. Seems there is room for furthere optimization by putting all duplicate code in a separate file/routine and call it in both places?

climbfuji (changes requested):

scm/etc/case_config/arm_sgp_summer_1997_A.nml
```diff
@@ -7,6 +7,7 @@ time_scheme = 1,
 runtime = 345600,
 n_itt_out = 1,
 n_levels = 64,
+n_itt_diag = 1,
```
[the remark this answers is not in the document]
grantfirl: Yes, but I'll remove it. I need to initiate an issue to clean up all case configuration files, especially once the dust settles on the new DEPHY format, so it is best to keep all that stuff together.

scm/src/gmtb_scm_time_integration.F90
grantfirl: I have mixed thoughts on this. Yes it is repeated. However, the part that is repeated in gmtb_scm.F90 is only exercised when using the leapfrog scheme for the forcing terms for since the first timestep since it must be a forward timestep. I am likely going to remove the option to use a different time-stepping scheme due to the KISS principle and I don't think it is necessary to keep it. Once that is removed, the repetition will be gone.

scm/src/GFS_typedefs.meta
grantfirl: Yep, thanks for catching this mistake. I'll push a commit once successfully recompiled.

grantfirl: Done.

scm/etc/case_config/arm_sgp_summer_1997_A.nml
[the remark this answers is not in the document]
grantfirl: Done.

scm/src/gmtb_scm_time_integration.F90
grantfirl: https://github.com/NCAR/ccpp-scm/issues/234

climbfuji:
Please let me know when you are done with the updates by requesting another review, and I'll take a look. Thanks!

grantfirl:
> Please let me know when you are done with the updates by requesting another review, and I'll take a look. Thanks!

There definitely need to be followup PR(s) to catch up to master, but I'd like to merge this before continuing on.

climbfuji:
@grantfirl I merged https://github.com/NCAR/ccpp-physics/pull/562 earlier today.

grantfirl:
> @grantfirl I merged [NCAR/ccpp-physics#562](https://github.com/NCAR/ccpp-physics/pull/562) earlier today.

Thanks. Did you want to review this PR again?

grantfirl:
I'll update the submodule pointer for ccpp-physics real quick.

grantfirl:
> I'll update the submodule pointer for ccpp-physics real quick.

Nevermind. Since that PR went into top of master, that will break things. I think that I'll just need to go ahead and update the SCM to the top of master.

climbfuji:
> > I'll update the submodule pointer for ccpp-physics real quick.
> 
> Nevermind. Since that PR went into top of master, that will break things. I think that I'll just need to go ahead and update the SCM to the top of master.

Yeah, no other choice, unfortunately. Cannot merge it somewhere in between.

grantfirl:
> > > I'll update the submodule pointer for ccpp-physics real quick.
> > 
> > 
> > Nevermind. Since that PR went into top of master, that will break things. I think that I'll just need to go ahead and update the SCM to the top of master.
> 
> Yeah, no other choice, unfortunately. Cannot merge it somewhere in between.

Totally fine. The update to top-of-master is long overdue anyway, so now is a good time.

grantfirl:
Replaced by #239

PR #226, opened 2021-02-10
Another specified surface flux input bugfix

grantfirl:
This PR fixes another bug reported by @egrell where the kinematic surface fluxes from existing specified surface flux cases were being read into a variable that holds surface fluxes in W m-2. This caused the surface fluxes used in the case to be set to their initial values (0) rather than those read from file.

This should hopefully be the last bugfix for existing specified surface flux cases introduced during the development to support the DEPHY format since I ran the same case using code before the DEPHY changes and, as of this PR, the output was identical.


llpcarson (approved):
approved

PR #246, opened 2021-03-30
Fix zero roughness length using DEPHY format with non-LSM ICs

### LINKED ISSUE #228: Surface friction in input_type=1
I am testing the use of the new DEPHY input format (input_type=1).  It appears that no surface friction is being calculated.  For sure it is not being output, variable ustar is identically zero.  This happens in the ARMCU_REF case with suite GSD_v1.  I have also tried it with a LLJ case I am developing, with both GSD_v1 and RRFS_v1beta suites.  The results are consistent with no surface friction.  It was being calculated correctly in the same case with the old input type.  I dug through the code some, and did not find an obvious problem.  The relevant attributes in the input file are:
 :z0 = 0.035 ;
 :surfaceType = "land" ;
 :surfaceForcing = "surfaceFlux" ;
 :surfaceForcingWind = "z0" ;

grantfirl:
Problem: For cases in the DEPHY format, the roughness length was not being properly initialized from the data file. It was read in properly, but in the case where the case data file does not contain model or LSM initial conditions, the values of the fractional landmask roughness length values (zorll, zorli, zorlw) were not being set from the one that is read in (zorlo). The logic assumed that input_zorl[l,i,w] would be set to a missing value (negative), but that does not happen in the read routine for the DEPHY format.

Solution: Modify the logic to set zorl[l,i,w] from zorlo if the input values are negative (missing) or zero (set to their initialization values).

Fixes #228

### DIFF
--- a/scm/src/gmtb_scm_type_defs.F90
+++ b/scm/src/gmtb_scm_type_defs.F90
@@ -1357,19 +1357,19 @@ module gmtb_scm_type_defs
         end if
       end if
       
-      if (scm_input%input_tsfcl < real_zero) then
+      if (scm_input%input_tsfcl <= real_zero) then
         physics%Sfcprop%tsfcl(i) = physics%Sfcprop%tsfco(i) !--- compute tsfcl from existing variables
       end if
       
-      if (scm_input%input_zorll < real_zero) then
+      if (scm_input%input_zorll <= real_zero) then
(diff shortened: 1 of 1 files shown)


grantfirl:
This was tested (on my Mac only) to not change the results of the supported cases and fixes the 'no surface friction' problem for the DEPHY cases reported by @wangevine.

climbfuji (approved):

PR #247, opened 2021-04-01
change to skeleton schemes

mzhangw:
Change to skeleton schemes for SCM exercise.

### DIFF
--- a/tutorial_files/add_new_scheme/fix_sys_bias_sfc.F90
+++ b/tutorial_files/add_new_scheme/fix_sys_bias_sfc.F90
@@ -20,7 +20,7 @@
 !> \section arg_table_fix_sys_bias_sfc_run Argument Table
 !! \htmlinclude fix_sys_bias_sfc_run.html
 !!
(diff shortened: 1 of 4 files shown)


ligiabernardet (changes requested):
I believe we do not want to modify the existing files, which have complete code. Instead, add two new files fix_sys_bias_sfc_skeleton.F90 and fix_sys_bias_sfc_skeleton.meta.

grantfirl:
> I believe we do not want to modify the existing files, which have complete code. Instead, add two new files fix_sys_bias_sfc_skeleton.F90 and fix_sys_bias_sfc_skeleton.meta.

I had the same question via email. We have two options:
1. Rename the skeleton files like you suggested.
2. Rename the completed files.

For subsequent exercises (after the add_a_scheme), students would simply edit the files that they copied into ccpp-physics, so that we don't also need to provide a 'skeleton' for the others, right?

grantfirl (approved):
Looks good. Thanks, Man.

PR #248, opened 2021-04-01
Fix Cheyenne consecutive builds (with different compilers)

grantfirl:
add deactivate at start of Cheyenne setup scripts to handle consecutive builds where compilers change; add GNU10 changes in CMakeLists.txt; upgrade to GNU10 in master (same as version 5 release)

This addresses an issue on Cheyenne where one could not build using Intel followed by a GNU build (for testing purposes). This has been tested to work on Cheyenne, switching from GNU to Intel without logging out/in.

Credit to @JulieSchramm for discovery and @llpcarson for solution.


grantfirl:
@JulieSchramm I also upgraded the GNU compiler to use 10.1 on Cheyenne. Hopefully, this will work for your RT stuff.

scm/etc/Cheyenne_setup_intel.sh
```diff
@@ -4,9 +4,10 @@ echo "Setting environment variables for SCM-CCPP on Cheyenne with icc/ifort"
 
 #load the modules in order to compile the GMTB SCM
 echo "Loading intel and netcdf modules..."
+deactivate
```
climbfuji, line 7: What happens if someone doesn't have the pylab environment loaded? Will this just do nothing?

grantfirl:
That's what happened when I tried it.

grantfirl:
> The simplest fix is to activate the NCAR python library at the top of the Cheyenne_setup* files (to avoid having to test if it is activated), then deactivate:
> 
> module load ncarenv
> ncar_pylib
> deactivate
> 
> The consecutive builds worked with this setup.

Thanks, Julie. I just tested this and it works for me too. It's a little weird to have to activate/deactivate/activate in the same setup script, but I can put a comment in there trying to explain it. I'll push a commit with this stuff added in a sec for you to review.

climbfuji (approved):

JulieSchramm (approved):
Looks good.

PR #261, opened 2021-05-18
changes related to ccpp-physics PR#627

grantfirl:
add GFS_radiation_surface scheme in ccpp_prebuild_config.py and all SDFs; add new variables in GFS_typedefs; use different sfcprop variables in scm_type_defs

Note: Checks on the variable 'nwat' were not transferred to the SCM version of the GFS_typedefs.F90 due to this variable being specific to FV3. Therefore, an additional difference exists between the SCM and FV3 versions of these files.


climbfuji:
ccpp-physics PR 627 was merged

climbfuji:
@grantfirl this can be merged now, correct?

grantfirl:
> @grantfirl this can be merged now, correct?

Probably, but I need to update submodule pointers and see if it compiles OK still. I'm also guessing that ccpp-physics main has moved on so that there will need to be more changes to build/compile, but hopefully it'll at least work with the commit of ccpp-physics where PR#627 was merged.

grantfirl:
@climbfuji As expected, this will no longer build/compile (even with the commit where #627 was merged because other stuff was merged into ccpp-physics beforehand) and needs to be updated. I'll update this pull request to work with the latest ccpp-physics main and run the new RTs.

grantfirl:
OK @climbfuji and @llpcarson, this PR should be good to go now. It was updated to work with the top of main ccpp-physics and ccpp-framework. It also passes @JulieSchramm RTs on Hera and Cheyenne. Please review when you get a chance. Most changes should be similar to recent fv3atm PRs.

climbfuji (approved):

PR #268, opened 2021-07-01
update GFS_typedefs.F90/meta to work with latest main ccpp-physics

grantfirl:
This updates the main branch of the SCM to work with the main branch of ccpp-physics as of 1 July, 2021.

This superficially follows https://github.com/NOAA-EMC/fv3atm/pull/310. That is, the changes to GFS_typedefs.F90/meta are brought over and the new tracer_types array now exists, but is set to all zeros. If GOCART is ever to be used with the SCM in the future, more of the fv3atm PR will probably need to be brought over and the tracer files reworked to support the different chemistry classifications.

### DIFF
--- a/scm/src/GFS_typedefs.F90
+++ b/scm/src/GFS_typedefs.F90
@@ -1064,8 +1064,8 @@ module GFS_typedefs
 
 !---cellular automata control parameters
     integer              :: nca             !< number of independent cellular automata
-    integer              :: nlives          !< cellular automata lifetime
-    integer              :: ncells          !< cellular automata finer grid
+    integer              :: tlives          !< cellular automata lifetime
+    integer              :: scells          !< cellular automata finer grid
     integer              :: nca_g           !< number of independent cellular automata
     integer              :: nlives_g        !< cellular automata lifetime
     integer              :: ncells_g        !< cellular automata finer grid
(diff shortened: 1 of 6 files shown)


climbfuji:
@grantfirl we just merged the GOCART integration code in ccpp-physics, do you want to make this a separate PR for the SCM?

grantfirl:
Dang it. No, I'll update this again.

grantfirl:
It looks like the GOCART changes in FV3 are extensive. It's probably going to take a while to figure out what needs to get transferred over. It looks like the tracers are completely different now? This was supposed to just be a quick update so that I could create the standard names PR, but no such luck, I guess.

climbfuji:
We can merge this PR as-is and you can do the GOCART changes as a follow-up?

grantfirl:
I think that the standard name changes (which is what this PR was supposed to facilitate) need to go on top of the main branches, so I don't see any point in merging as-is.

grantfirl:
After looking a bit more at the GOCART changes in fv3atm, I think it will be OK (for now) to hard-code the new tracer_types array as all zeros rather than copy over some of the logic for the changed field_tables etc. found in the fv3atm GOCART PR. If anyone ever wants GOCART to work with the SCM, the rest of the changes can be revisited at that time, but I have not heard anything in regards to this. If I don't need to make the SCM work with GOCART, the update is suddenly much easier and I just need to bring over a subset of the changes in fv3atm.

grantfirl:
These code changes passed the SCM RTs (B4B identical with previous main).

[rt_summary63168.txt](https://github.com/NCAR/ccpp-scm/files/6750109/rt_summary63168.txt)

grantfirl:
@climbuji Please review when you get a chance. Most changes should be familiar. If there is disagreement about whether true GOCART functionality is needed in the SCM at this point, we should discuss.

climbfuji (approved):

PR #277, opened 2021-10-15
Add support for NSSL cloud microphysics scheme (CCPP)

MicroTed:
See CCPP pull request for NSSL scheme features (https://github.com/NCAR/ccpp-physics/pull/761)
Related PR for FV3: https://github.com/NOAA-EMC/fv3atm/pull/472

Adds suites, namelists, and tracer config files for NSSL (with/without hail/CCN)

Namelist options:

nssl_cccn - CCCN background CCN concentration. Default: 0.6e9 (per m^3 at STP)
nssl_alphah - graupel shape parameter, default = 0.
nssl_alphahl - hail shape parameter, default = 1.
nssl_hail_on - hail activation flag, default = .false.
nssl_ccn_on - CCN prediction flag, default = .true.
nssl_invertccn - store CCN as number of activated instead of number remaining, default = .true. (avoids need for specified boundary conditions)

Example test run:
./run_scm.py -c twpice -n input_GSD_v1nssl_nohailnoccn.nml -t tracers_gsd_nssl_nohail_noccn.txt -s SCM_GSD_v1nssl

Also made a change to netcdf output to set the initial date info as integers (matching the native data type). This also solves a type mismatch issue with the analysis python program. (Error passing date info as floats instead of integers.)


MicroTed:
Updated to 28 Dec. 2021 working version of SCM+CCPP (newer CCPP seems to have unsupported variables)

grantfirl:
> Updated to 28 Dec. 2021 working version of SCM+CCPP (newer CCPP seems to have unsupported variables)

@MicroTed The SCM has been updated to the latest ccpp-physics/framework. This means that you should be able to update your ccpp-physics PR and this one and it should work.

MicroTed:
OK, cool, I'll try updating. The last that I worked on fv3atm (week or two ago), it seemed to be good with the then-latest CCPP, as well.

MicroTed:
This and the CCPP PR have been updated to current main

ccpp/physics_namelists/input_RRFS_v1nssl.nml
```diff
@@ -0,0 +1,95 @@
...
+    iau_delthrs = 6
+    iau_inc_files = ''
+    iaufhrs = 30
+    icloud_bl = 1
+    ico2 = 2
+    iems = 1
+    imfdeepcnv = -1
+    imfshalcnv = -1
+    imp_physics = 17
+    nssl_ccn_on = .false.
```
grantfirl, line 36: Is this namelist supposed to have nssl_ccn_on and nssl_hail_on = F? The other one (input_RRFS_v1_nssl_nohailnoccn.nml) already has these as F.

MicroTed: Good catch, I probably forgot I had changed it during testing. Since the default for the suite is

"SCM_RRFS_v1nssl":"input_RRFS_v1nssl_nohailnoccn.nml"

Then I could either remove input_RRFS_v1nssl.nml or change it to turn on both hail and CCN as an example. Although I suppose the point of SCM is not to get the 'best' result but rather to check that the scheme is working.  Generally, I'd like to set things up for the intended use/needs of SCM. Hail was turned off here because the scheme probably would not produce any, and it's not an output variable anyway.

grantfirl (approved):

PR #279, opened 2021-10-19
Correct spelling of janic in CCPP standard names, must be janjic

climbfuji:
As the title says, correct spelling of janic in CCPP standard names, must be janjic (affects `GFS_typedefs.meta` only). This PR must be merged with/after https://github.com/NCAR/ccpp-physics/pull/752.


grantfirl:
Addressed in #282

PR #283, opened 2021-11-16
Adding more diagnostics for help with cloud fraction scheme(s)

gthompsnWRF:
Six new variables - 2D only - were added to support cloud fraction schemes to diagnose impacts of explicit cloud condensate versus impacts from cloud fraction.  There is a corresponding branch with the same name in the ccpp-physics repo to include the new calculations together with a new/updated cloud fraction scheme.


scm/src/GFS_typedefs.F90
```diff
@@ -6893,7 +6911,7 @@ subroutine diag_phys_zero (Diag, Model, linit, iauwindow_center)
     Diag%dku = zero
 
 ! max hourly diagnostics
-    Diag%refl_10cm   = zero
+    Diag%refl_10cm   = -35.
```
climbfuji, line 6914: This change makes sense to me, but I want to check with @grantfirl if there was a particular reason to reset `Diag%refl_10cm` to zero instead of `-35.`

climbfuji (commented):
The changes add the required variables to the GFS data structure and the CCPP metadata, but I am wondering if there are additional steps needed to output these variables in the single column model? Or has this not been done, because you were using the UFS in your tests?

scm/src/GFS_typedefs.F90
gthompsnWRF: As FYI to this point. Many people think dbz=zero means no precip. That is not true.  Very light drizzle occurs all the time with dbz of about -5 to -10 values.  The -35 value is considered (approximately) the lowest sensitivity level of the USA's NEXRAD radars.  Cloud radars are sensitive down to even smaller values, but the frequency/wavelength of NEXRAD is 10cm, hence why I was so specific in its name.

gthompsnWRF:
> The changes add the required variables to the GFS data structure and the CCPP metadata, but I am wondering if there are additional steps needed to output these variables in the single column model? Or has this not been done, because you were using the UFS in your tests?

That's correct.  I added these 6 new variables to GFS_diagnostics.F90 to get them for UFS and not really sure what/where to place them in SCM - unless they go in scm_output.F90.

scm/src/GFS_typedefs.F90
grantfirl: No, @gthompsnWRF reasoning makes perfect sense. This was only zero in the SCM version of GFS_typedefs to match FV3.

grantfirl:
The new variables need to be added to scm_output.F90 only. I can do this for you @gthompsnWRF (I'll branch off of your PR branch, make modifications, and submit a PR back into your branch so that it will be reflected here once accepted by you). It's pretty straightforward, but there could be a complication or two depending on what is being held in these variables. The Diag DDT in GFS_typedefs holds variables that can be instantaneous, accumulated, etc. and that needs to be accounted for in how they're written out in scm_output.F90.

grantfirl:
This PR is superceded by #285, so closing this one.

PR #285, opened 2021-11-19
Wrapper for #283, fix RT bug, add radiation output interval

grantfirl:
- add another radiation time interval for output

Add an output interval that corresponds to when either SW or LW are called

- fix interstitial variable output on radiation timesteps

This fixes a problem identified by @siwei-noaa that some interstitial variables calculated prior to calling radiation are being output on every timestep (including all the timesteps when they are zero) instead of only on timesteps when radiation is called.

- fix rt.sh not correctly reporting differing output values

The way that the cmp command return value was being checked led to false positives (reporting identical output when not actually identical)

- includes all changes in #283 that were not previously merged as a result of catching up with fv3atm


grantfirl:
Changes RT baseline due to changing output frequency for a few variables

grantfirl:
Since this PR includes changes in #283, I'm closing that one in favor of this one.

grantfirl:
Regression tests run on Mac but have differing baselines as expected.

climbfuji (approved):
Looks good. I couldn't see from the code how the output will look like. Do the rad variables have the same time dimension, but missing values at the timesteps when radiation is not run? Or do they have a second time dimension?

grantfirl:
> Looks good. I couldn't see from the code how the output will look like. Do the rad variables have the same time dimension, but missing values at the timesteps when radiation is not run? Or do they have a second time dimension?

A while ago, there was only one time dimension associated with output and for any values that were not valid on that time interval (e.g. radiation-related ones), the data was output as missing. Several people didn't like this, so prior to this PR, within the last year or so, there were 4: one for instantaneous output, one for the diagnostic interval where values are averaged over the previous interval, one for instantaneous values when SW rad is called, and one for instantaneous values when LW rad is called. This PR adds a fifth time interval for when any rad is called, either SW or LW.

siwei-noaa:
Grant, thank you for working on this. I just tested with the latest code, and the output looks correct.

PR #292, opened 2022-01-18
 Change netcdf definition of init_year etc. from float to integer

MicroTed:
The python program scm_analysis.py fails on the datetime.datetime function because it expects init_year, init_month etc. to be integers. Since these variables are declared as integers in the SCM code, this PR changes the netcdf output of the init_year, etc. from float to integer. The python read then recognizes the variables as integers and runs correctly.

An alternate temporay solution is to convert to integer in the script, e.g.,

from
year.append(nc_fid.variables['init_year'][:])
to
year.append(int(nc_fid.variables['init_year'][:]))


grantfirl:
@MicroTed Thanks. I'd like to pull this into another branch where I've fixed a couple of other output-related bugs and submit a PR for all of them together. Is that OK with you?

MicroTed:
@grantfirl That's is certainly fine with me. Let me know if I need to do anything on my end

grantfirl:
These changes are included in #293 now, so closing this one.

PR #293, opened 2022-01-20
Fix output-related bugs

grantfirl:
- move increment of itt_out to output_append to be consistent with other output time indices
- add option to force writing instantaneous output (outside of normal cadence, e.g. initial conditions)
- call output_append at every timestep and return if no output should be written during current timestep
- fix bug when n_rad == 1 and/or when radiation is not called (missing value fills)
- fix type of initial date variables (thanks @MicroTed))
- write out appropriate missing value depending on output datatype in NetCDF_def_var (thanks @MicroTed)

This should cause failed SCM RTs due to changed output files.

### DIFF
--- a/scm/etc/scripts/scm_analysis.py
+++ b/scm/etc/scripts/scm_analysis.py
@@ -42,6 +42,13 @@ except (AttributeError):
 def print_progress(n_complete, n_total):
     print(str(n_complete) + ' of ' + str(n_total) + ' complete: (' + str(100.0*n_complete/float(n_total)) + '%)')
 
+def replace_fill_with_nan(nc_ds, var_name, var, group):
+    raw_data = nc_ds.variables[var_name][:]
+    raw_data[raw_data == nc_fid.variables[var_name]._FillValue] = np.nan
+    var.append(raw_data)
+    group.append(var_name)
+    return [var, group]
+
 #set up command line argument parser to read in name of config file to use
 parser = argparse.ArgumentParser()
 parser.add_argument('config', help='configuration file for CCPP SCM analysis', nargs=1)
@@ -123,8 +130,6 @@ for filename in glob.glob(os.path.join(dir, '*.nml')):
     nml = f90nml.read(filename)
     if nml['case_config']:
         case_name = nml['case_config']['case_name']
-
-
 
 #initialize lists for output variables
 year = []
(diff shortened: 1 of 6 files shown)


grantfirl:
Replaces #292

scm/src/scm_output.F90
```diff
@@ -392,35 +394,43 @@ subroutine output_init_diag(ncid, time_inst_id, time_diag_id, time_rad_id, hor_d
...
-  allocate(dummy_1D(scm_state%n_cols), dummy_2d(scm_state%n_cols, scm_state%n_levels))
+  integer :: ncid, var_id
 
   !> \section output_append_alg Algorithm
   !! @{
-
+  if (.not. (mod(scm_state%itt, scm_state%n_itt_out)==0 .or. &
+            physics%Model%lsswr .or. physics%Model%lslwr .or. &
+            mod(scm_state%itt,physics%Model%nszero) == 0 .or. &
+            (PRESENT(force) .and. force))) then
```
climbfuji, line 412: Note: This is not safe. The Fortran programming standard does NOT say that `PRESENT(force)` will be executed before testing if `force` is true, nor does it require the compiler to skip the test for `force` being true, even if the first expression evaluates to false.

You need to adjust the logic to first test for present only, and inside that if block test if force is true.

For example
```
if (present(force)) then
  force_local = force
else
  force_local = .false.
endif
```
and then use `force_local` instead of `force`.

scm/src/scm_output.F90
```diff
@@ -392,35 +394,43 @@ subroutine output_init_diag(ncid, time_inst_id, time_diag_id, time_rad_id, hor_d
...
+    return
+  end if
   !> - Open the file.
   CALL CHECK(NF90_OPEN(PATH=TRIM(scm_state%output_dir)//"/"//TRIM(scm_state%output_file)//".nc",MODE=NF90_WRITE,NCID=ncid))
   
   !> - Append all of the variables to the file.
-  CALL CHECK(NF90_INQ_VARID(NCID=ncid,NAME="time_inst",VARID=var_id))
-  CALL CHECK(NF90_PUT_VAR(NCID=ncid,VARID=var_id,VALUES=scm_state%model_time,START=(/ scm_state%itt_out /)))
+  if (mod(scm_state%itt, scm_state%n_itt_out)==0 .or. (PRESENT(force) .and. force)) then
```
climbfuji, line 419: Same here.

climbfuji (changes requested):

scm/src/scm_utils.F90
```diff
@@ -660,13 +660,21 @@ subroutine NetCDF_def_var(ncid, var_name, var_type, desc, unit, varid, dims)
...
-      CALL CHECK(NF90_DEF_VAR(NCID=ncid,NAME=var_name,XTYPE=var_type,DIMIDS=dims,VARID=varid))
+      call CHECK(NF90_DEF_VAR(NCID=ncid,NAME=var_name,XTYPE=var_type,DIMIDS=dims,VARID=varid))
     else
-      CALL CHECK(NF90_DEF_VAR(NCID=ncid,NAME=var_name,XTYPE=var_type,VARID=varid))
+      call CHECK(NF90_DEF_VAR(NCID=ncid,NAME=var_name,XTYPE=var_type,VARID=varid))
+    end if
+    call CHECK(NF90_PUT_ATT(NCID=ncid,VARID=varid,NAME="description",VALUES=desc))
+    call CHECK(NF90_PUT_ATT(NCID=ncid,VARID=varid,NAME="units",VALUES=unit))
+    
+    if (var_type == NF90_FLOAT) then
```
mkavulich, line 670: Is there any chance that variables could be NF90_DOUBLE?

mkavulich:
Looks good aside from the issue Dom brought up and my one comment.

Can you document what tests were run for this PR?

scm/etc/scripts/scm_analysis.py
```diff
@@ -785,7 +750,7 @@ def print_progress(n_complete, n_total):
     date_swrad.append(np.array([initial_date + datetime.timedelta(seconds=int(s)) for s in time_swrad[-1]]))
     date_lwrad.append(np.array([initial_date + datetime.timedelta(seconds=int(s)) for s in time_lwrad[-1]]))
     date_rad.append(np.array([initial_date + datetime.timedelta(seconds=int(s)) for s in time_rad[-1]]))
-
+
```
climbfuji, line 753:
```suggestion

```

climbfuji (approved):

grantfirl:
Thanks @climbfuji and @mkavulich for the review and suggestions, particularly pointing out the optional argument problem that I introduced. For testing, I tried several different combinations of n_itt_out and n_itt_diag in the case configuration files. There had been a bug where the instantaneous variables were trying to be written out one too many times with some combinations of these values. This is definitely fixed. I ran the SCM RTs and the results were as-expected: all tests pass with some differences to baselines. I checked the differences using nccmp manually and the only difference was a change in data type from float->integer for the expected variables. This is the expected result. I also plotted some results from the TWPICE case using all supported suites and everything looked good except for some of the tendency plots were plotting the FILL values as if they were good values. Some changes to the scm_analysis.py script fixed that and the plots are now good to go for now.

MicroTed:
> except for some of the tendency plots were plotting the FILL values as if they were good values

I noticed this issue in recent tests -- thanks for fixing it!

grantfirl:
fixes #294

MicroTed:
@grantfirl 
 By the way, the NaN conversion is clever, but gives me this error (anconda, Python 3.8.5 and 3.8.12):

   raise ValueError("Axis limits cannot be NaN or Inf")
ValueError: Axis limits cannot be NaN or Inf

It works if I set the values to zero instead of np.nan. I've seen that NaN values are supposed to be skipped in the plotting, but maybe it doesn't like having all the values be nan?

Issue #298, opened 2022-02-08
Problem to install a shapely python module


wzzheng90:
I got a problem when I set the environment variables for CCPP-SCM on Hera with icc/ifort as follows:

    % source scm/etc/Hera_setup_intel.csh

"Hera_setup_intel.csh" needs to install a shapely python module:
  
      pip install --index-url http://anaconda.rdhpcs.noaa.gov/simple --trusted-host [anaconda.rdhpcs.noaa.gov](http://anaconda.rdhpcs.noaa.gov/) shapely --user

The error message shows:
        Retrying (Retry(total=4, connect=None, read=None, redirect=None)) after connection broken by 'ConnectTimeoutError(<pip._vendor.requests.packages.urllib3.connection.HTTPConnection object at 0x7f8d26951828>, 'Connection to [anaconda.rdhpcs.noaa.gov](http://anaconda.rdhpcs.noaa.gov/) timed out. (connect timeout=15)')': /simple/shapely/

How to fix this problem?   Thanks!!

Weizhong

grantfirl:
Hi Weizhong, I'll look into it and make any changes to the Hera setup script as necessary.

wzzheng90:
@grantfirl :  Really appreciate!!

grantfirl:
@wzzheng90 Take a look at #299. Would you try to modify your Hera_setup_intel.csh file and see if this solution works for you?

wzzheng90:
@grantfirl :   Try this command:   source scm/etc/Hera_setup_intel.csh   and the message shows as follows:

Setting environment variables for CCPP-SCM on Hera with icc/ifort
Loading intel and netcdf modules...
Setting CC/CXX/FC environment variables
Setting NCEPLIBS environment variables
Loading cmake
Loading the SCM python environment
.: Permission denied.
conda: Command not found.

wzzheng90:
All right.  I just saw you changed it again "source /scratch1/BMC/gmtb/SCM_anaconda/etc/profile.d/conda.csh".
Now type:   source scm/etc/Hera_setup_intel.csh   and no error message appears:

Setting environment variables for CCPP-SCM on Hera with icc/ifort
Loading intel and netcdf modules...
Setting CC/CXX/FC environment variables
Setting NCEPLIBS environment variables
Loading cmake
Loading the SCM python environment

=======  
However, when I compile the code, error messages show as follows:

[...]

PR #299, opened 2022-02-09
update Hera setup scripts

grantfirl:
Fixes https://github.com/NCAR/ccpp-scm/issues/298

Hera sysadmins changed how python is managed on Hera: https://rdhpcs-common-docs.rdhpcs.noaa.gov/wiki/index.php/Anaconda#Migrating_from_the_old_.22RDHPCS_mirror.22_installation_to_the_Anaconda_installation

Specifically, installing shapely, configobj, and netCDF4 using the method of grabbing the modules from the RDHPCS mirror does not work any more. According to their documentation, the preferred method for managing python is for users to install their own version of Miniconda and use that, which is what I did on Hera.

- I installed miniconda in /scratch1/BMC/gmtb/SCM_anaconda on Hera 
- create new "pyccpp" environment to maintain for CCPP SCM users (we can install whatever packages that we want in that environment in the future); this eliminates the need to check for installed packages and install them if missing -- if future users have a python package problem on Hera, it will be on SCM maintainers to update the pyccpp environment.

Note that UFS is going to use the hpc-stack repo for managing dependencies on supported platforms, which should also affect these files. I tried to follow their instructions for using hpc-stack modules on Hera and ran into a runtime error when running the SCM executable (something about not being able to find the sz.o dynamic library, potentially related to szip, hdf5, and/or use of netCDF in the code).

### DIFF
--- a/scm/etc/Hera_setup_intel.csh
+++ b/scm/etc/Hera_setup_intel.csh
@@ -28,54 +28,6 @@ setenv CMAKE_CXX_COMPILER icpc
 setenv CMAKE_Fortran_COMPILER ifort
 setenv CMAKE_Platform hera.intel
 
-echo "Loading the anaconda python distribution"
-module use -a /contrib/anaconda/modulefiles
-module load anaconda/anaconda3-4.4.0
-
-#install f90nml for the local user
-
-#check to see if f90nml is installed locally
-echo "Checking if f90nml python module is installed"
-python -c "import f90nml"
-
-if ( $? != 0 ) then
-	echo "Not found; installing f90nml"
-	pip install -e git://github.com/marshallward/f90nml.git@v0.19\#egg=f90nml --user
-else
-	echo "f90nml is installed"
-endif
-
-#install shapely for the local user
-
-#check to see if shapely is installed locally
-echo "Checking if shapely python module is installed"
-python -c "import shapely"
-
-if ( $? != 0 ) then
-	echo "Not found; installing shapely"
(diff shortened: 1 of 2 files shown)


climbfuji (approved):
Changes look good and fix the problem reported in https://github.com/NCAR/ccpp-scm/issues/298.

siwei-noaa:
May I know if only one step is required to use this new Hera setup script: source ../etc/Hera_setup_intel.csh? When I tried to use it, I got the message:
Setting environment variables for CCPP-SCM on Hera with icc/ifort
Loading intel and netcdf modules...
Setting CC/CXX/FC environment variables
Setting NCEPLIBS environment variables
Loading cmake
Loading the SCM python environment
usage: conda [-h] [-V] command ...

conda is a tool for managing and deploying applications, environments and packages.

Options:
positional arguments:

...

conda commands available from other packages:
  env
Could not find conda environment: pyccpp
You can list all discoverable environments with `conda info --envs`.

With this, I can compiled the SCM successfully, but I could not run it. The error message is:
Traceback (most recent call last):
  File "/scratch2/BMC/gsd-fv3-test/she/ccpp-scm/scm/bin/./run_scm.py", line 5, in <module>
    import f90nml
ModuleNotFoundError: No module named 'f90nml'

Thank you,
Siwei

siwei-noaa:
Please forget my previous message. I figured out the reason. 
I have installed my Anaconda, and it loads automatically when I log in Hera. In this way, I cannot load your SCM python environment from your Anaconda (`source Hera_setup_intel.csh` could not be finished correctly). After unloading my Anaconda environment, I can source `Hera_setup_intel.csh` without any issue. 

Thank you,
Siwei

PR #317, opened 2022-05-20
V6 release sdfs/namelists

grantfirl:
This pull request accomplishes the following tasks:

- add SDFs and namelists for release v6 supported suites
- update suite defaults for supported suites
- update code to work with latest main
- fix bug in SCM_forcing for DEPHY-format cases
- update CMakeLists.txt to specify which suites to compile with (can specify -DCCPP_SUITES=ALL, -DCCPP_SUITES=SCM_GFS_v16,....[comma-separated list], or if -DCCPP_SUITES is not included, the supported suites listed in suite_info.py are compiled)


dustinswales (commented):
@grantfirl 
This looks good to me, just one small question on the nmls.

scm/etc/scripts/mod_forcing.py
```diff
@@ -0,0 +1,423 @@
...
+parser.add_argument('--micro',            help='add microphysics tendency to forcing terms', action='store_true', default=False)
+parser.add_argument('--ogwd',             help='add orographic gravity wave drag tendency to forcing terms', action='store_true', default=False)
+parser.add_argument('--cgwd',             help='add convective gravity wave drag tendency to forcing terms', action='store_true', default=False)
+parser.add_argument('-v', '--verbose',    help='once: set logging level to debug; twice: set logging level to debug '\
+                                               'and write log to file', action='count', default=0)
+
+###############################################################################
+# Functions and subroutines                                                   #
+###############################################################################
+def setup_logging(verbose):
```
climbfuji, line 46: Is this a duplicate of what is in `run_scm.py`? If so, there is some room for improvement by putting that logic in a shared Python module.

scm/etc/scripts/mod_forcing.py
```diff
@@ -0,0 +1,423 @@
...
+    """Sets up the logging module."""
+    # print out debug messages (logs and output from subprocesses) if verbose argument is set
+    if verbose==2:
+        LOG_LEVEL = logging.DEBUG
+    elif verbose==1:
+        LOG_LEVEL = logging.INFO
+    else:
+        LOG_LEVEL = logging.WARNING
+    LOG_FILE = 'multi_run_scm.log'
```
climbfuji, line 55: This still says `multi_run_scm.log`

climbfuji (approved):

mkavulich (approved):
Looks good, but a question for my own benefit: what is the deal with all of these "*_ps" suites? They seem to be identical to the main suite but missing some surface schemes?

ccpp/physics_namelists/input_GFS_v17_p8.nml
```diff
@@ -0,0 +1,139 @@
...
+  nlives       = 12
+  nseed        = 1
+  nfracseed    = 0.5
+  nthresh      = 18
+  ca_trigger   = .true.
+  nspinup      = 1
+  iseed_ca     = 12345
+  lndp_type    = 0
+  n_var_lndp   = 0
+/
```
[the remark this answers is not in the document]
grantfirl: Sure, I don't see the harm in that. I'll add it.

scm/etc/scripts/mod_forcing.py
grantfirl: Yes. This script is a prototype. This can be fixed when it gets made more robust/useful and is no longer a prototype.

scm/etc/scripts/mod_forcing.py
grantfirl: Thanks, good catch.

ccpp/physics_namelists/input_GFS_v17_p8.nml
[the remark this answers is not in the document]
grantfirl: See https://github.com/NCAR/ccpp-scm/pull/317/commits/c0dbebd327c0377685f7c86e70b85ea18043e994

scm/etc/scripts/mod_forcing.py
grantfirl: See https://github.com/NCAR/ccpp-scm/pull/317/commits/c0dbebd327c0377685f7c86e70b85ea18043e994

scm/src/suite_info.py
```diff
@@ -57,7 +61,19 @@ def timestep(self, value):
...
-  
+suite_list.append(suite('SCM_csawmg',            'tracers_csawmg.txt',                   'input_csawmg.nml',                  600.0, 1800.0, False))
+
+def main():
+    
+    #print supported suites separated by commas
+    suite_string = ''
+    for s in suite_list:
+        if s._supported:
+            suite_string += s._name + ',' + s._name + '_ps' + ','
```
dustinswales, line 74: @grantfirl 
This behaves differently than before.
Before all suites in "suite_list" were created, but now only supported suites in "suite_list" are created. 
Also, isn't this also trying to create a "_ps" counterpart suite for each supported suite?

dustinswales (commented):
@grantfirl
I think there is an issue with which SDFs are being created.

scm/src/suite_info.py
grantfirl: @dustinswales Yes, this works differently than before. Previously, the CMakeLists.txt file told cmake to create a list of all suites to compile no matter what. Now, which suites get compiled depends on whether -DCCPP_SUITES is given to cmake. The old behavior should be retrieved by specifying -DCCPP_SUITES=ALL. If one omits -DCCPP_SUITES, CMakeLists.txt now calls this script, which returns a string of suites to compile, specifically, those where the `supported` attribute is true.

grantfirl: The number of suites has been growing, leading to really long compilation times. This is supposed to help that, and now it works more similarly to UFS too.

dustinswales: @grantfirl
Thanks for the explanation. I will update the tutorial to reflect this change.

grantfirl: Also, this isn't creating any SDFs. It's only building a list of existing SDFs to actually create caps for and compile.

dustinswales (approved):

scm/src/suite_info.py
dustinswales: You are correct. I mean't "which SDFs are used"

grantfirl:
> > Looks good, but a question for my own benefit: what is the deal with all of these "*_ps" suites? They seem to be identical to the main suite but missing some surface schemes?
> 
> You've correctly identified the difference! They are suites with the surface schemes removed in favor of using data for surface fluxes and a special CCPP scheme that backs out some variables needed by PBL schemes from the specified surface fluxes. Many SCM cases like to use specified surface fluxes since it removes a rather large degree of freedom for folks interested in how the atmospheric physics (only) responds to a given set of forcing.
> 
> The way the code is set up is for the run script to use the right SDF (the special *_ps ones) if someone requests to run a case with specified surface fluxes. This is transparent to the user who doesn't need to explicitly tell the code to use the special suite. I'm not in love with how I implemented this and there are certainly ways to do this so as not to have to maintain two different SDFs...

I figured I'd write this down for posterity (mainly for my own benefit) before I forgot... 

I think that maybe a cleaner way of doing this would be to have one SDF for each suite, but to have it differ from the FV3 SDFs by having 2 extra groups -- one that gets called when surface schemes should be called and one that gets called when surface fluxes are specified. This way, it could just be handled in the SCM Fortran by using the CCPP API group calling functionality rather than all of the run script nonsense with duplicate SDFs that we're using now.

PR #321, opened 2022-06-14
update SCM code to use latest CCPP main branches

grantfirl:
This updates the SCM to use the latest CCPP physics/framework main branches and addresses some issues pointed out be @egrell :
- fixes some mistakes in the Users' Guide
- add the timestep and output intervals to the run script output
- fix a typo in the run script


ligiabernardet:
I incorrectly labelled his PR as needed for CCPP v6. I see this pertains to main. So I have not removed the CCPP v6 label. Sorry for the confusion.

scm/src/GFS_typedefs.F90
```diff
@@ -2380,15 +2408,18 @@ subroutine sfcprop_create (Sfcprop, IM, Model)
        end if
 
     end if
+       allocate (Sfcprop%rmol   (IM ))
```
mkavulich, line 2411: Reduce the indentation here as these lines are not in an `if` statement.

mkavulich (approved):
I didn't spot any errors here, just had one minor comment.

scm/src/GFS_typedefs.F90
grantfirl: I noticed this, but the same misindentation is in the FV3 version too (and brought over via Meld). Correcting it in the SCM introduces a difference that I'd rather not deal with. Perhaps if I remember, I can fix it in both places the next time we have cause to change GFS_typedefs.

PR #322, opened 2022-06-15
Update CCPP commit hashes; fix output bug; update ccpp_prebuild_config.py

grantfirl:
This has been tested to build and run on my M1 Mac with the latest CCPP release branches.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,7 +5,7 @@
 [submodule "ccpp-physics"]
 	path = ccpp/physics
 	url = https://github.com/NCAR/ccpp-physics
-	branch = main
+	branch = release/public-v6
 [submodule "CMakeModules"]
 	path = CMakeModules
 	url = https://github.com/noaa-emc/CMakeModules
--- a/ccpp/config/ccpp_prebuild_config.py
+++ b/ccpp/config/ccpp_prebuild_config.py
@@ -263,7 +263,9 @@ SUITES_DIR = 'ccpp/suites'
 
 # Directory where to write static API to
 STATIC_API_DIR = 'scm/src/'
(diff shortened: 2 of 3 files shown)


ligiabernardet (commented):
Script `contrib/get_all_static_data.sh` is downloading v5 dataset
`wget https://github.com/NCAR/ccpp-scm/releases/download/v5.1.0/${file}.tar.gz`

Need to create a v6 dataset and change the script.

ligiabernardet (approved):
It is okay to merge this PR. A few follow up changes will be needed:
- I was able to build and run on Hera using `bash`. When I used `tcsh` I could not build. I am not sure if the problem is on my end.
- In file `scm/src/run_scm.py`, I see `DEFAULT_SUITE = 'SCM_GFS_v15p2'`. How about changing the default to the `SCM_GFS_v16` suite?
- `scm/etc/scripts/plot_configs/*.ini` files still refer to old suites

grantfirl:
> It is okay to merge this PR. A few follow up changes will be needed:
> 
> * I was able to build and run on Hera using `bash`. When I used `tcsh` I could not build. I am not sure if the problem is on my end.
> * In file `scm/src/run_scm.py`, I see `DEFAULT_SUITE = 'SCM_GFS_v15p2'`. How about changing the default to the `SCM_GFS_v16` suite?
> * `scm/etc/scripts/plot_configs/*.ini` files still refer to old suites

@ligiabernardet What kind of error were you getting with tcsh on Hera? I can build/run just fine with tcsh on Hera, BTW.

I will change the default suite in run_scm.py. I can change suites in the plot_configs, although they are mostly there for example. Perhaps I'll pick a couple out to update the suites.

PR #325, opened 2022-06-21
Release/public v6 gjf

grantfirl:
- conditionally write tendency terms to output file (only if the index in the big tendency array is >= 1, i.e. used)
-- this replaces the previous function of always initializing the tendency variables in the output file whether they were used or not in the run, resulting in arrays of missing values
- the scm_analysis.py script was modified to treat these variables as optional variables in the output; if they're not present for a given output file, they're filled in with zeros (not NaNs or -999) because they are often plotted together with other tendencies, and it is better to see 0 tendencies than -999 or something
- reinstate conv_tendencies plot for tutorial


scm/etc/scripts/plot_configs/twpice_all_suites.ini
```diff
@@ -1,5 +1,5 @@
-scm_datasets = output_twpice_SCM_GFS_v16/output.nc, output_twpice_SCM_HRRR/output.nc, output_twpice_SCM_RAP/output.nc, output_twpice_SCM_RRFS_v1beta/output.nc, output_twpice_SCM_WoFS_v0/output.nc
-scm_datasets_labels = GFSv16, HRRR, RAP, RRFSv1beta, WoFS_v0
+scm_datasets = output_twpice_SCM_GFS_v16/output.nc, output_twpice_SCM_GFS_v17_p8/output.nc, output_twpice_SCM_RAP/output.nc, output_twpice_SCM_RRFS_v1beta/output.nc, output_twpice_SCM_WoFS_v0/output.nc, output_twpice_SCM_HRRR/output.nc
+scm_datasets_labels = GFSv16, GFSv167p8, RAP, RRFSv1b, WoFSv0, HRRR
```
dustinswales, line 2: Fix typo in label name

dustinswales:
@grantfirl
Looks great!
Just a small typo fix

scm/etc/scripts/plot_configs/twpice_all_suites.ini
grantfirl: Thanks

dustinswales (approved):

PR #326, opened 2022-06-21
Some Tutorial and Docker-related updates for v6

mkavulich:
- Updating Dockerfile for latest release
   - Adding myself as a maintainer
   - restoring the old NCEPLIBS build script as a stopgap before we upgrade to newest hpc-stack or spack
   - Call new data retrieval script
   - Add environment variables that are needed for running so user doesn't need to set them manually
 - Updating run script
   - Remove extra debug layer to avoid having normal messages show up as warnings
   - Update "copy_outdir" function to be able to work from any directory, not just run directory (this prevents errors in docker container)

Tests run: Used this branch (modified slightly to avoid chicken-egg problem) to build dockerfile and run tutorial case successfully with resulting docker image (can be downloaded from Dockerhub: `docker pull dtcenter/ccpp-scm:v6.0.0-beta`)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 FROM dtcenter/common-community-container:gnu9
 
-MAINTAINER Michelle Harrold <harrold@ucar.edu> or Grant Firl <grantf@ucar.edu>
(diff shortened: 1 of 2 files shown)


grantfirl (approved):
This looks good. Thanks, @mkavulich

PR #329, opened 2022-06-23
Final v6 Users' Guide updates

grantfirl:
This should finish up the edits needed in the UG, unless reviewers disagree.


ligiabernardet (commented):
@DomHeinzeller Your affiliation the User's Guide  is currently listed as JCSDA and DTC. I would like to change that to CU/CIRES at NOAA/GSL and DTC, since your contribution was not funded by JCSDA. We can also add a footnote that says Currently affiliated with JCSDA. LMK what you prefer.

ligiabernardet:
As @mkavulich said yesterday, we need an update to scm/contrib/get_thompson_tables.sh because the name of the file to download does not match the available file (tar vs tar.gz).

ligiabernardet:
It is okay but not necessary to run the following command since we are not supporting suites that use MG mp.
`./ contrib / get_mg_inccn_data .sh`

ligiabernardet:
In some places we provide commands for both bash and csh, while in other places we only provide commands for bash. E.g.:
`cmake [- DCMAKE_BUILD_TYPE ...] ../ src 2 >&1 | tee log.cmake`
`make -j4 2 >&1 | tee log.make`
Do we want to address this?

ligiabernardet:
Questions about the input datasets:

- CCPP-SCM format: _Each file has two dimensions (time and levels)._ This is true for the supported cases, so perhaps this sentence is fine. But there are cases in the repo with additional dimensions, including soil, right?
- DEPHY format: Section 5.1.3 says: _The initialization and forcing data for each case is stored in a NetCDF (version 4) file within the ccpp-scm/scm/data/processed_case_input directory._ I don't think so. I think the files are in the DEPHY repo and must be staged in that directory prior to using, correct? Also, does the user need to tell the SCM which format is being used? Or does it automatically recognize it?

ligiabernardet:
Post-release: consider removing gjf from script names in `/scm/etc/scripts` (and make corresponding change in UG)

ligiabernardet (changes requested):
Doc is looking great! Just a few minor comments.
Also need a final change in one of the .sh scripts

ligiabernardet:
The script is issue is addressed in https://github.com/NCAR/ccpp-scm/pull/330

grantfirl:
@ligiabernardet I'll update this later today.

dustinswales:
> Questions about the input datasets:
> 
> * CCPP-SCM format: _Each file has two dimensions (time and levels)._ This is true for the supported cases, so perhaps this sentence is fine. But there are cases in the repo with additional dimensions, including soil, right?
> * DEPHY format: Section 5.1.3 says: _The initialization and forcing data for each case is stored in a NetCDF (version 4) file within the ccpp-scm/scm/data/processed_case_input directory._ I don't think so. I think the files are in the DEPHY repo and must be staged in that directory prior to using, correct? Also, does the user need to tell the SCM which format is being used? Or does it automatically recognize it?

@ligiabernardet There is a new case_configuration variable, input_type, to tell the SCM which input file format.

grantfirl:
> @grantfirl @dustinswales As far as Python modules, the User's Guide talks about the need to have f90nml. But the tutorial instructions also mention the need for xarray, dask, netCDF4, bottleneck, matplotlib, configobj, shapely, pandas, and scipy. Are those other Python modules only needed for the tutorial (that is, for plotting)? Or should they also be mentioned in the User's Guide?

Only f90nml is needed to build and run. The others are for plotting. There are more Python packages used by the UFS IC functionality, but those are mentioned in section 5.5.

grantfirl:
> In some places we provide commands for both bash and csh, while in other places we only provide commands for bash. E.g.: `cmake [- DCMAKE_BUILD_TYPE ...] ../ src 2 >&1 | tee log.cmake` `make -j4 2 >&1 | tee log.make` Do we want to address this?

I'm not an expert in bash/tcsh differences. If someone wants to go through with a finetooth comb looking for commands that won't execute with one shell or the other, please do so and suggest changes.

grantfirl:
> Questions about the input datasets:
> 
> * CCPP-SCM format: _Each file has two dimensions (time and levels)._ This is true for the supported cases, so perhaps this sentence is fine. But there are cases in the repo with additional dimensions, including soil, right?
> * DEPHY format: Section 5.1.3 says: _The initialization and forcing data for each case is stored in a NetCDF (version 4) file within the ccpp-scm/scm/data/processed_case_input directory._ I don't think so. I think the files are in the DEPHY repo and must be staged in that directory prior to using, correct? Also, does the user need to tell the SCM which format is being used? Or does it automatically recognize it?

Ya, the DEPHY-SCM cases are not in the repo. I changed the text to reflect that and added some instructions if anyone wants to clone the repo and try those cases. As @dustinswales said, there is a namelist variable in the case configuration namelist that denotes what kind of file to expect. It is not automatic.

grantfirl:
@ligiabernardet You will need to rereview since you requested changes. I addressed Dom's affiliation, answered your question about Python modules (although didn't make any UG changes), removed the MG data fetch command from the UG,  added some more info about running with cases from the DEPHY-SCM repo, and changed the default suite in the run script (I thought I had already done this, but I hadn't -- shouldn't affect tutorial, etc.).

ligiabernardet (approved):

PR #330, opened 2022-06-24
A few final v6 changes

mkavulich:
- Add missing environment variable to Dockerfile
 - Update all static data files to .gz for consistency and smaller file size


grantfirl (approved):
Thanks @mkavulich

PR #331, opened 2022-06-25
Fix environment variable declarations for Cheyenne GNU bash

mkavulich:
The Cheyenne GNU environment setup file had some extra spaces in variable assignments, causing errors. This fixes that issue.


ligiabernardet (approved):
Thank you for these changes.

PR #344, opened 2022-08-18
Update SCM to latest CCPP physics/framework main branches

grantfirl:
- use w3emc library instead of w3nco
- update CCPP submodule pointers
- add LICENSE


grantfirl:
@mkavulich @dustinswales Could one of you review these changes when you get a chance so that the SCM is up-to-date with the latest CCPP please?

dustinswales (approved):
@grantfirl I have no comments. All of these changes look good to me.

LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright 2017, NOAA, UCAR/NCAR, CU/CIRES, CSU/CIRA
```
mkavulich, line 1:
```suggestion
Copyright 2022, NOAA, UCAR/NCAR, CU/CIRES, CSU/CIRA 
```

scm/src/GFS_typedefs.F90
```diff
@@ -3362,7 +3371,11 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
 !  max and min lon and lat for critical relative humidity
     integer :: max_lon=5000, max_lat=2000, min_lon=192, min_lat=94
     real(kind=kind_phys) :: rhcmax = 0.9999999               !< max critical rel. hum.
+#ifdef SINGLE_PREC
+    real(kind=kind_phys) :: huge   = 9.9692099683868690E30  !  NetCDF float FillValue
```
mkavulich, line 3375: Just curious, do you know the reason for these seemingly random values? It seems like a different value is used in fv3gfs (https://github.com/NOAA-EMC/fv3gfs/blob/main/sorc/fv3gfs.fd/FV3/gfsphysics/GFS_layer/GFS_typedefs.F90#L16)

mkavulich (changes requested):
Just one question and a suggested change (make copyright date current), rest looks good.

scm/src/GFS_typedefs.F90
grantfirl: They must be related to the largest value that can fit into this particular datatype for some machine, perhaps WCOSS. The values in GFS_typedefs.F90 are the same in this repo and NOAA-EMC/fv3atm. I don't know why we would need to compare to a much older version of GFS_typedefs in the fv3gfs repo.

grantfirl:
@mkavulich I committed your suggested change to the license file. You'll need to re-review in order to remove the merge block.

mkavulich (approved):

PR #348, opened 2022-10-18
Updates to testing in SCM

dustinswales:
The PR contains two changes:
- Here we introduce continuous integration tests to the ccpp-scm using GitHub workflows. There are two tests: The first CI test is a simple workflow that runs the ccpp_prebuild_config.py for the ccpp-scm, using the accompanying ccpp-physics. The second CI test goes further and runs the full SCM workflow, downloading and building dependencies and running the SCM regression-tests using the GNU compilers.
- There is also a change to the run_scm.py script when running with multiple cases. The input lists have been replaced by a list of python dictionaries, one for each run configuration.


.github/workflows/ci_build_scm_gnu9.4.yml
```diff
@@ -0,0 +1,113 @@
...
+      run: |
+        conda install --yes -c conda-forge f90nml
+        conda install --yes -c conda-forge netCDF4
+
+    - name: Set compilers
+      run: |
+        echo "FC=gfortran-9" >> $GITHUB_ENV
+        echo "CC=gcc-9"      >> $GITHUB_ENV
+
+    - name: Install bacio library v2.4.1
```
grantfirl, line 47: Do we really have to install these libraries every time? There's no way to do something similar as Python with `uses:`? Seems like a waste of cycles, but if that's what has to happen, then so be it. It'd be great if there was container support for CI.

dustinswales: @grantfirl Good point. I will look into this further.

dustinswales: @grantfirl Turns out you can cache installs in these workflows.[ I added a few more steps to save/cache some libraries.](https://github.com/dustinswales/ccpp-scm/blob/update_multirun/.github/workflows/ci_build_scm_gnu9.4.yml#L58)

grantfirl:
@dustinswales I'll approve once https://github.com/dustinswales/ccpp-scm/pull/6 is merged.

grantfirl (approved):

PR #351, opened 2022-11-03
Updates for ccpp-physics ufs-dev PR#10

grantfirl:
- update GFS_typedefs.F90/meta to have latest changes for external land component support


grantfirl:
@dustinswales The CI scripts are running through the RTs, but not actually running rt.sh, right? So we're not comparing against any baselines ATM prior to merging PRs? Or, should we be doing this manually?

dustinswales:
@grantfirl You are correct. 
This piece, data storage, needs to be sussed out. The CI won't be able to upload to hera/cheyenne, at least I don't think that's feasible? maybe it is.
The CI needs an additional step to download the RT baselines, which need to be stored somewhere at the DTC, and to run ./rt.sh instead of run_scm.py.

dustinswales (approved):

Issue #234, opened 2021-03-02
Remove option for forcing time-stepping scheme


grantfirl:
The filtered-leapfrog scheme adds complexity and is not regularly tested. Remove to follow KISS principle and simplify codebase.

PR #362, opened 2023-01-17
Address issue 234

dustinswales:
Address issue #234

### DIFF
--- a/scm/src/scm.F90
+++ b/scm/src/scm.F90
@@ -82,8 +82,6 @@ subroutine scm_main_sub()
   select case(scm_state%time_scheme)
     case(1)
       cdata_time_index = 1
-    case(2)
-      cdata_time_index = 2
     case default
       cdata_time_index = 2
   end select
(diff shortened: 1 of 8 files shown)


scm/src/scm.F90
```diff
@@ -82,8 +82,6 @@ subroutine scm_main_sub()
   select case(scm_state%time_scheme)
```
grantfirl, line 82: if `time_scheme` only has one possible value (1), why don't we get rid of it altogether and remove this select/case statement?

scm/src/scm_input.F90
```diff
@@ -46,7 +46,7 @@ subroutine get_config_nml(scm_state)
   integer              :: n_snow   !< number of model snow levels (currently only 3 supported)
   integer              :: n_columns !< number of columns to use
   integer              :: n_time_levels
```
grantfirl, line 48: `n_time_levels` controls the size of some state arrays, but if we're getting rid of the only time scheme that required more than one time level, we can get rid of this too. Of course, it would involve going through the code and removing a dimension from the state variables. Is the simplification worth the work?

PR #364, opened 2023-01-23
Add UFS replay + DEPHY v1

dustinswales:
This PR contains two innovations. 

The first is the functionality to create SCM cases using standard output from the UFS weather model. The changes contained here are built on-top of https://github.com/grantfirl/ccpp-scm/tree/UFS_forcing_20210203/scm. The dynamic tendencies to drive the SCM are derived in UFS_IC_generator.py. Then, along with the necessary initial conditions, these derived forcing terms are stored in DEPHYv1 format. 

The second addition is the ability for the SCM to use input data in the [DEPHYv1 format],(https://docs.google.com/document/d/1eAWY-ELL5Ua6a9WIsv4ODHmLXvfgla5TNQAuAwNASo0/edit). This gives the SCM access to all of the cases in the [GdR-DEPHY repository](https://github.com/GdR-DEPHY/DEPHY-SCM).

Documentation changes needed.

Also included: 
- small change to the error checking routine to print the offending command in case of error.
- new CI test for DEPHY v1 GdR cases (This points to their HEAD, so it will let use know if (when) the format in that repository decides to change)


dustinswales:
Closed in favor of #370 and #372

PR #366, opened 2023-02-28
Update SCM for NCAR PR #956 and UFS #35

dustinswales:
This brings the SCM up-to-date with the HEAD of [NCAR:main](https://github.com/NCAR/ccpp-physics).
Changes include: [UWM:ufs/dev #35](https://github.com/ufs-community/ccpp-physics/pull/35) and [NCAR:main #956](https://github.com/NCAR/ccpp-physics/pull/956)

### DIFF
--- a/scm/src/GFS_typedefs.F90
+++ b/scm/src/GFS_typedefs.F90
@@ -713,6 +713,12 @@ module GFS_typedefs
     integer              :: nrcm            !< second dimension of random number stream for RAS
     integer              :: iflip           !< iflip - is not the same as flipv
     integer              :: isol            !< use prescribed solar constant
+                                            !< 0  => fixed value=1366.0\f$W/m^2\f$(old standard)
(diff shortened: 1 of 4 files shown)


dustinswales:
@grantfirl All tests passed. I think good to merge (see below)

Baselines are different for a few p8 cases, which is expected since there are changes to the Thompson MP scheme, but not all the p8 cases? I wonder if [this change](https://github.com/ufs-community/ccpp-physics/pull/35/files) doesn't get triggered/exercised by all cases in the SCM RTs with Thompson MP enabled? (These are itsy-bitsy cases we are running).

grantfirl (approved):
Ya, I'm guessing that the arm_sgp_summer_1997_A case and TWPICE cases have different baselines, but that BOMEX, ASTEX, and LASSO cases do not? That would be consistent with the changes only manifesting themselves in deep convection.

PR #370, opened 2023-03-22
Feature DEPHY v1

dustinswales:
This PR adds the ability to use [DEPHY v1](https://github.com/GdR-DEPHY/DEPHY-SCM) forcing files with the SCM.

The changes are isolated to the scm input/output/util modules to accommodate the new format (see [DEPHY v1 doc](https://docs.google.com/document/d/1eAWY-ELL5Ua6a9WIsv4ODHmLXvfgla5TNQAuAwNASo0/edit))
- From DEPHY v0(?) -> v1 there was a change in the format to remove the spatial indexing. For example, 2D data format changed from [x, y, z, time] -> [z,time].
- Variable name changes.
- New input argument (char) for the "check" subroutine. Calls to check were modified to print this descriptor when there is an error. (There were variable name changes from DEPHY v0 -> v1, and knowing what was failing was impossible w/o this added input to check)

There is a new case configuration file and CI test to exercise the SCM's compatibility with the DEPHY repository, [ci_run_scm_DEPHY.yml](https://github.com/dustinswales/ccpp-scm/blob/feature_dephy_v1/.github/workflows/ci_run_scm_DEPHY.yml#L129). (If they change their format, yet again, this test will break. At which time we can decide to either update the SCM to use the latest DEPHY format or just use the last working hash in the DEPHY repository.)

This replaces part of #364


dustinswales:
@grantfirl All tests pass. No changes to baselines.

scm/src/scm_input.F90
```diff
@@ -770,6 +770,16 @@ subroutine get_case_init(scm_state, scm_input)
   scm_input%input_vegsrc   = input_vegsrc
   
   scm_input%input_slmsk    = input_slmsk
+  scm_input%input_canopy   = input_canopy
```
grantfirl, line 773: These changes came from #364, but are not needed for this PR, right?

scm/src/scm_input.F90
```diff
@@ -922,82 +932,225 @@ subroutine get_case_init_DEPHY(scm_state, scm_input)
...
-  real(kind=sp), allocatable :: input_qi(:,:,:,:) !< IC ice water specific humidity profile (kg kg^-1)
-  real(kind=sp), allocatable :: input_rv(:,:,:,:) !< IC water vapor mixing ratio profile (kg kg^-1)
-  real(kind=sp), allocatable :: input_rt(:,:,:,:) !< IC total water mixing ratio profile (kg kg^-1)
-  real(kind=sp), allocatable :: input_rl(:,:,:,:) !< IC liquid water mixing ratio profile (kg kg^-1)
-  real(kind=sp), allocatable :: input_ri(:,:,:,:) !< IC ice water mixing ratio profile (kg kg^-1)
-  real(kind=sp), allocatable :: input_rh(:,:,:,:) !< IC relative humidity profile (%)
-  real(kind=sp), allocatable :: input_tke(:,:,:,:) !< IC TKE profile (m^2 s^-2)
+  real(kind=dp), allocatable :: input_lat(:) !< column latitude (deg)
+  real(kind=dp), allocatable :: input_lon(:) !< column longitude (deg)
+  !real(kind=dp), allocatable :: input_zorog(:)  ! surface forcing:
```
grantfirl, line 955: Still need this line?

scm/src/scm_input.F90
```diff
@@ -922,82 +932,225 @@ subroutine get_case_init_DEPHY(scm_state, scm_input)
...
-  real(kind=sp), allocatable :: input_force_temp_nudging(:,:,:,:)
-  real(kind=sp), allocatable :: input_force_theta_nudging(:,:,:,:)
-  real(kind=sp), allocatable :: input_force_thetal_nudging(:,:,:,:)
-  real(kind=sp), allocatable :: input_force_qv_nudging(:,:,:,:)
-  real(kind=sp), allocatable :: input_force_qt_nudging(:,:,:,:)
-  real(kind=sp), allocatable :: input_force_rv_nudging(:,:,:,:)
-  real(kind=sp), allocatable :: input_force_rt_nudging(:,:,:,:)
-  real(kind=sp), allocatable :: input_force_u_nudging(:,:,:,:)
-  real(kind=sp), allocatable :: input_force_v_nudging(:,:,:,:)
+  real(kind=sp), allocatable :: input_force_pres_surf(:) !< forcing surface pressure (Pa)
```
grantfirl, line 1117: In the description, you note a change in dimensions in the DEPHY v1 file, but I must have interpreted your description wrong. You said "from DEPHY v0(?) -> v1 there was a change in the format from [x, y, z] -> [column, z]". This looks like there is every only data for 1 column and the dimensions for 1D arrays are [time], and the dimensions for 2D arrays are [z, time]. Is that right?

scm/src/scm_input.F90
```diff
@@ -1006,293 +1159,602 @@ subroutine get_case_init_DEPHY(scm_state, scm_input)
...
   end if
   ierr = NF90_INQ_DIMID(ncid,"nice",varID)
   if(ierr /= NF90_NOERR) then
     input_n_ice = missing_ice_layers
   else
-    call check(NF90_INQUIRE_DIMENSION(ncid, varID, tmpName, input_n_ice))
+    call check(NF90_INQUIRE_DIMENSION(ncid, varID, tmpName, input_n_ice),"nf90_inq_dim(nice)")
   end if  
   
+  model_ics=.false.
```
grantfirl, line 1204: Does this need to stay? I thought that the UFS forcing generator script was outputting in DEPHY format, and thus reading in this file, you might have the model "restart"-type data. Isn't this disabling that?

scm/src/scm_input.F90
dustinswales: Yes. I cherry picked a few files from #364 to create this PR, which was a bit too much for one PR IMO. I didn't want to remove the UFS-replay changes manually, so it just left this stuff in there. A new PR for UFS-replay is imminent anyways.

scm/src/scm_input.F90
dustinswales: No. Removed

scm/src/scm_input.F90
dustinswales: No. Good catch. Removed. I replaced this by querying the DEPHY global attribute surface_forcing_lsm.

grantfirl:
> > Looks good, @dustinswales . Just a few questions.
> 
> @grantfirl Thanks for your review.
> 
> There are [some documentation changes](https://github.com/NCAR/ccpp-scm/commit/c148a2ac7b3edeb8b01b9d59baf9338e788bca25) that I need to add. I don't have a latex editor on my work machine to update main.pdf, and these changes are on my home computer. So I will update the PDF later. Question. Do we update the date on the Users Doc?

We've only updated the date for releases in the past, but it's also fine if you'd rather do that now too.

PR #372, opened 2023-04-04
Feature: UFS Replay

dustinswales:
The first is the functionality to create SCM cases using standard output from the UFS weather model. The changes contained here are built on-top of https://github.com/grantfirl/ccpp-scm/tree/UFS_forcing_20210203/scm. The dynamic tendencies to drive the SCM are derived in UFS_IC_generator.py. Then, along with the necessary initial conditions, these derived forcing terms are stored in DEPHYv1 format.


dustinswales:
@grantfirl Documentation added, section 5.5. I think this is ready to go otherwise.

mkavulich:
@dustinswales I have taken a brief look at the code but was waiting to test it myself as I see you are still making changes to the CI tests. Do you think this PR is stable enough to start a thorough review?

dustinswales:
@mkavulich Thanks! Yes, it's ready for review.
The CI test is working, well mostly. It does two things. A) Creates a UFS-replay case from staged UWM RT data on the DTC FTP website and then B) Builds and runs SCM with the UFS_replay case. It's failing at the very end of the last step.

Maybe this is all too much anyways? Since what we really need is to be alerted to UWM input/output changes that break the case generation script (we have another CI test that does the full workflow for DEPHY formatted input, which is what comes out of the case generation script). The functionality of this test is only as useful if the data from the UWM is somewhat up-to-date.

dustinswales:
@grantfirl The new CI test to create the UFS-replay cases from staged UWM RT data is working, and passing.

scm/doc/TechGuide/chap_cases.tex
```diff
@@ -132,7 +132,7 @@ \section{Included Cases}
 \item UFS initial conditions for 38.1 N, 98.5 W (central Kansas) for 00Z on Oct. 3, 2016 with Noah variables on the C96 FV3 grid (\execout{fv3\_model\_point\_noah.nc})
 \item UFS initial conditions for 38.1 N, 98.5 W (central Kansas) for 00Z on Oct. 3, 2016 with NoahMP variables on the C96 FV3 grid (\execout{fv3\_model\_point\_noahmp.nc})
 \end{itemize}
-See \ref{sec:UFS ICs} for information on how to generate these files for other locations and dates, given appropriate UFS Atmosphere initial conditions.
+See \ref{sec:UFSreplay} for information on how to generate these files for other locations and dates, given appropriate UFS Atmosphere initial conditions.
```
grantfirl, line 135: "... give appropriate UFS Atmosphere **output**."?

scm/doc/TechGuide/chap_cases.tex
```diff
@@ -214,29 +235,130 @@ \section{Using UFS Initial Conditions}
...
+(-lonl LON_1 LON_2 -latl LAT_1 LAT_2 -nens NENSMEMBERS |
+-lons [LON_LIST] -lats [LAT_LIST])
+[-dt TIMESTEP] [-cres C_RES] [-sdf SUITE] [-sc] [-near]
+\end{lstlisting}
+
+Mandatory arguments:
+\begin{enumerate}
+\item \exec{-{}-dir (-d)}: path to UFS Regression Test output
+\item \exec{-{}-case\_name (-n)}: name of cases
+\item Either: (see examples bewlow)
```
grantfirl, line 270: typo: "bewlow"

scm/etc/scripts/UFS_forcing_ensemble_generator.py
```diff
@@ -0,0 +1,193 @@
...
+parser.add_argument('-near', '--use_nearest',   help='flag to indicate using the nearest UFS history file gridpoint, no regridding',action='store_true')
+
+###############################################################################
+# Main program
+###############################################################################
+def main():
+    # Get command line arguments
+    args  = parser.parse_args()
+
+    if (not args.dir):
```
grantfirl, line 37: This shouldn't be necessary due to the `required=True` for the `dir` argument.

scm/etc/scripts/UFS_forcing_ensemble_generator.py
```diff
@@ -0,0 +1,193 @@
...
+    # Create inputs to SCM
+    case_list    = ""
+    case_list_nf = ""
+    count = 0
+    run_list = []
+    for pt in range(0,npts):
+        # Call UFS_IC_generator.py
+        case_name     = args.case_name +"_n" + str(pt).zfill(3)
+        file_scminput = "../../data/processed_case_input/"+case_name+"_SCM_driver.nc"
+        com = "./UFS_IC_generator.py -l " +str(lons[pt]) + " " + str(lats[pt]) + \
```
grantfirl, line 126: Is this slow? Would it benefit from using multiprocessing, e.g. a solution like the `execute` function in the run script?

grantfirl: This refers to using the `os.system(com)` serially, below.

grantfirl (approved):
This looks good to me. I'm OK with it merged as-is and any comments addressed later. Thanks for pushing it across the finish line.

scm/etc/scripts/UFS_IC_generator.py
```diff
@@ -694,1391 +1304,1436 @@ def get_UFS_surface_data(dir, tile, i, j, old_chgres):
...
+    for t in range(n_files):
+        #
+        dtime_sec = (stateNATIVE["time"][t+1] - stateNATIVE["time"][t])*sec_in_hr
+        #
+        pres_adv[t,:]   = stateNATIVE["p_lay"][t,:]
+        pres_i_adv[t,:] = stateNATIVE["p_lev"][t,:]
+        #
+        tend_total[0,:] = stateREGRID["t_lay"][t+1,:] - stateREGRID["t_lay"][t,:]
+        tend_remap[0,:] = stateREGRID["t_lay"][t,:]   - stateNATIVE["t_lay"][t,:]
+        dtdt_adv[t,:]   = (tend_total[0,:] - tend_remap[0,:]) / dtime_sec
```
grantfirl, line 1887: IMO, this method should be revised in the (near) future. Calculating tendencies this way includes the effect of physics already, which we want the physics in the SCM to represent. I think the "right" way to calculate advective tendencies, if we don't want to rely on having either the non-physics tendency terms or the all-physics tendency terms as "special" diagnostics (which most runs will not save due to space constraints), is to actually use the output winds with neighboring profiles of T, q with the upwind differencing scheme to derive them. I'll go ahead and start an issue so that we don't forget to do this.

dustinswales: You are correct, I think of this as the quick and easier(dirtier) way.
I'm out of my lane when it comes to numerical recipes for computing advection, but it's not clear to me how using neighboring points gets around the implicit physics influence on the state? (I need to learn more)

scm/etc/scripts/UFS_forcing_ensemble_generator.py
dustinswales: Good point. I will remove.

PR #377, opened 2023-04-07
UFS-dev PR#57

grantfirl:
- changes to typedefs to work with ufs/dev PR#57: https://github.com/NCAR/ccpp-physics/pull/1010

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,10 +4,8 @@
 	branch = main
 [submodule "ccpp-physics"]
 	path = ccpp/physics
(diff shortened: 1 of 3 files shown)


grantfirl:
I ran RTs on my Mac by going through a similar procedure to what is in [ci_run_scm_rts.yml](https://github.com/NCAR/ccpp-scm/blob/main/.github/workflows/ci_run_scm_rts.yml) and they passed as expected, using baselines from the current main top-of-branch.

PR #380, opened 2023-05-04
UFS-dev PR#56

grantfirl:
Changes analogous to https://github.com/NCAR/fv3atm/pull/93 to work with https://github.com/NCAR/ccpp-physics/pull/1013


grantfirl:
@dustinswales We couldn't update the baseline artifacts on the DTC FTP server from the last PR: https://github.com/NCAR/ccpp-scm/pull/376, so you might need to do that before RTs run here are valid.

PR #381, opened 2023-05-09
UFS-dev PR#64

grantfirl:
Contains changes analogous to https://github.com/NCAR/fv3atm/pull/94 for working with https://github.com/NCAR/ccpp-physics/pull/1015

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,8 +4,10 @@
 	branch = main
 [submodule "ccpp-physics"]
 	path = ccpp/physics
-	url = https://github.com/NCAR/ccpp-physics
-	branch = main
+	#url = https://github.com/NCAR/ccpp-physics
+	#branch = main
+        url = https://github.com/grantfirl/ccpp-physics
+        branch = ufs-dev-PR64
 [submodule "CMakeModules"]
 	path = CMakeModules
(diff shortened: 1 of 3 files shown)


dustinswales:
@grantfirl Looks like the RAP twpice test differs, [which is expected](https://github.com/NCAR/ufs-weather-model/pull/97), but not the other RAP tests?

grantfirl:
> @grantfirl Looks like the RAP twpice test differs, [which is expected](https://github.com/NCAR/ufs-weather-model/pull/97), but not the other RAP tests?

Ya, it is a bit surprising that the arm_sgp_summer_1997_A case does not also differ with the RAP suite because it is the only other case that features deep convection and would have a chance at triggering the GF code changes. But, I'm assuming that whatever convection occurs in the SGP case is not strong enough to trigger the code changes, perhaps?

grantfirl:
> That's what I suspect. I don't have a sense for the forcing in the all of the test cases or how this change manifests itself within the convection. I know we are only running 0.1x of these cases for our tests, and TWPICE is the longest test, which maybe is why it's triggering the convection?

Ya, that's yet another reason for the ARM case to not differ that I always forget (that we're only running 10% of the case model time), since I think the deep convection probably doesn't happen right away.

PR #386, opened 2023-06-08
Add link to the PDF guide in the readme

zmoon:
So that on GitHub, you can easily get to it.


grantfirl:
I've added a link in the README.md in #399 . I'd rather not rely on a third-party service for the PDF.

PR #387, opened 2023-06-14
UFS-dev PR#26

grantfirl:
analogous to https://github.com/NOAA-EMC/fv3atm/pull/606

Note that although these changes allow the SCM to compile and run through the field campaign cases, more work will need to be done in the input and UFS column simulator script to work with a lake point.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,10 +4,8 @@
 	branch = main
 [submodule "ccpp-physics"]
 	path = ccpp/physics
-	#url = https://github.com/NCAR/ccpp-physics
-	#branch = main
-        url = https://github.com/grantfirl/ccpp-physics
-        branch = ufs-dev-PR26
+	url = https://github.com/NCAR/ccpp-physics
+	branch = main
 [submodule "CMakeModules"]
 	path = CMakeModules
 	url = https://github.com/noaa-emc/CMakeModules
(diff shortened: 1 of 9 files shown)


grantfirl:
@mkavulich One more to review for ufs/dev PR#26 ...

mkavulich (approved):
@grantfirl Sorry about that, approved.

Random question if you get a chance: is there a reason the SCM maintains its own duplicate list of constants instead of using physcons.F90 from ccpp-physics? I've never noticed that's the case before this PR.

grantfirl:
> @grantfirl Sorry about that, approved.
> 
> Random question if you get a chance: is there a reason the SCM maintains its own duplicate list of constants instead of using physcons.F90 from ccpp-physics? I've never noticed that's the case before this PR.

I think that we (Dom and others) that physcons.F90 doesn't belong in the CCPP since it is a host-specific file. For some reason, this hasn't been removed from ccpp-physics, but the SCM was coded as-if it doesn't exist.

dustinswales:
> > @grantfirl Sorry about that, approved.
> > Random question if you get a chance: is there a reason the SCM maintains its own duplicate list of constants instead of using physcons.F90 from ccpp-physics? I've never noticed that's the case before this PR.
> 
> I think that we (Dom and others) that physcons.F90 doesn't belong in the CCPP since it is a host-specific file. For some reason, this hasn't been removed from ccpp-physics, but the SCM was coded as-if it doesn't exist.

I second that emotion.

PR #397, opened 2023-08-29
Fix input for LSM ICs in UFS replay. Optimize i/o step.

dustinswales:
This PR fixes a bug in how the LSM Initial conditions are set for UFS replay. 
Addresses differences seen in https://github.com/NCAR/ccpp-scm/discussions/369


dustinswales:
@grantfirl @mkavulich 
The replay CI test is failing because there was a [variable name change](https://github.com/NCAR/ccpp-scm/actions/runs/6017448830/job/16323658652#step:9:31) from the staged UFS output on the DTC server and what was being used by in #369. I will look into whether we need to update our data, or revert the name change in the code.

dustinswales:
@XiaSun-Atmos You can use it now if you want, just as long as you're aware that it may change slightly during the review process.

grantfirl:
@dustinswales I'd like to work on this one. As far as you are aware, this is still needed, right? What kind of testing was done?

XiaSun-Atmos:
I used this hotfix for the SCM run with active LSM for the CAPE paper.

dustinswales:
@grantfirl This PR is superseded by #434 (which has this fix and others needed by the replay)

PR #400, opened 2023-09-13
UFS-dev PR#78

grantfirl:
Updates corresponding to: https://github.com/NCAR/fv3atm/pull/101

I added noahmptable.tbl to the physics_input_data release asset so that it gets copied to the run directory and won't cause an error when running a suite that uses NoahMP.

Edit: This also includes a bugfix identified by @hertneky in scm_input.F90, changing one of the `if` tests to check for `kinematic` rather than `surfaceFlux`.


mkavulich (approved):

grantfirl:
Thanks to @hertneky for finding the bug in scm_input.F90.

PR #403, opened 2023-11-02
TESTING ONLY: Debugging CI tests requires active PR.

dustinswales:


grantfirl:
We can close this since CI was fixed in https://github.com/NCAR/ccpp-scm/pull/408

PR #407, opened 2023-11-08
Use spack-stack-1.5.1 on Hera with Intel (sh so far only)

climbfuji:
Switch to spack-stack-1.5.1 - so far only Hera/Intel and bash. I was able to compile the scm with this, didn't run the tests yet.

The only thing that is missing is doxygen, which is optional for building the documentation.


grantfirl:
@climbfuji @mkavulich This is effectively replaced by https://github.com/NCAR/ccpp-scm/pull/406 at this point, correct?

climbfuji:
Yes, closing. Thanks!

PR #435, opened 2024-03-06
Continuous Integration with Nvidia compiler

dustinswales:
This PR introduces support to use the SCM with NVidia HPC fortran compiler (nvfortran).
The CI tests build and run a set of regression tests using the SCM_RAP suite for two configurations, one without and one with GPU support. GPU runtime functionality is not tested, just the build step.


dustinswales:
@grantfirl This one is good to go whenever we want to add it!

dustinswales:
Note that the Debug builds are failing (mersenne_twister conflict) with Nvidia, so I have them off in the yaml file.

.github/workflows/ci_build_scm_ubuntu_22.04_nvidia.yml
```diff
@@ -0,0 +1,264 @@
...
+      run: |
+        cd ${SCM_ROOT}
+        ./contrib/get_all_static_data.sh
+        ./contrib/get_thompson_tables.sh
+
+    - name: Run SCM RTs (w/o GPU)
+      if: contains(matrix.enable-gpu-acc, 'False')
+      run: |
+        cd ${SCM_ROOT}/scm/bin
+        ./run_scm.py --file /home/runner/work/ccpp-scm/ccpp-scm/test/rt_test_cases.py --runtime_mult 0.1 -v
```
dustinswales, line 264: @grantfirl @mkavulich @scrasmussen 
This is incomplete, it only runs the RTs, it doesn't download baselines and compare them. [We should add these steps in](https://github.com/NCAR/ccpp-scm/blob/main/.github/workflows/ci_run_scm_rts.yml#L178) and store a copy of the Nvidia Baselines on the DTC FTP server, analogous to the GNU based ones we have there. 
So instead of rt-baselines-Release.zip and rt-baselines-Debug.zip on the DTC FTP server we will have:
rt-baselines-Debug.GNU.zip, rt-baselines-Release.GNU.zip, rt-baselines-Debug.NVHPC.zip, rt-baselines-Release.NVHPC.zip.

dustinswales: Sorry, no rt-baselines-Debug.NVHPC.zip, this test is off because it fails due to a naming conflict (mersenne_twister).

grantfirl:
Combined into #459

mkavulich:
@dustinswales @grantfirl It appears as if the CI test of this workflow is failing in the main branch. Since it seems like it was working on the last commit here, perhaps there was a problem with the merged PR?

PR #461, opened 2024-04-24
Add Simplified UFS_case_gen.py

grantfirl:
@dustinswales @scrasmussen @hertneky @ligiabernardet 

This is a PR to add a new, simplified version of UFS_IC_generator.py. Rather than modify the existing UFS_IC_generator.py, I've just created a new script named UFS_case_gen.py. We can change this if necessary. 

The script pulls both the initial condition profiles and forcing from UFS history files. The only reliance on the "true" initial conditions is for the surface and oro variables. This means a couple of things:

1. The case is created for the column on the diagnostic, Gaussian grid point, not the actual, native FV3 grid point. This eliminates any need for regridding in the horizontal, but it immediately creates a difference between the UFS and SCM runs. This should be acceptable, IMO.
2. The SCM initial conditions are taken from the first UFS history file, which is after the first time step. This should also be an acceptable compromise, given differences already coming from point 1.

The forcing section is technically split by "method". I have placeholder logic for calculating forcing using the original "gold standard" method and the existing method, that I call the "total time tendency" method. Right now, the new forcing method is the only one implemented in the script to show us how much simpler it can be. 

The advective forcing is calculated using centered differences in the horizontal (using either 1 or 2 halo points (i.e. a 3x3 or 5x5 grid with the column of interest in the center). The vertical differences are calculated using upstream differences, using a moving-averaged smoothed vertical velocity. Right now, the script is still outputting the "total" advective tendencies like the old UFS forcing, but it should be possible to output the horizontal advective tendencies and the smoothed vertical velocity profile for forcing, which is a preferable way to force SCMs.

I've tested this script using the same (global) data that was staged for testing the old script. I'm sure that more testing needs to be done, especially with output from a regional configuration, but I've chosen many random points and run the SCM for 24h cases. I only ran it for 24h because the UFS data that was used only has 12-hourly data for 1 day. Comparing only mean profiles of the state variables, the agreement is reasonably good. This is encouraging since we only have 12-hourly data! This method should be more accurate with better time resolution of the UFS data.

Known problems:
The forcing method is not supported very close to the poles.

### DIFF
--- a/scm/etc/scripts/UFS_case_gen.py
+++ b/scm/etc/scripts/UFS_case_gen.py
@@ -1,0 +1,2105 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+import fnmatch
+import logging
+from netCDF4 import Dataset
+import numpy as np
+from shapely.geometry import Point, Polygon
+import copy
+import math
+from datetime import datetime, timedelta
+from pylab import plot, ylim, xlim, show, xlabel, ylabel, grid
+
+###############################################################################
+# Global settings                                                             #
+###############################################################################
+
+#Physical constants
+earth_radius = 6371000.0 #m
+rdgas        = 287.05
+rvgas        = 461.50
+cp           = 1004.6
+zvir         = rvgas/rdgas - 1.
+rocp         = rdgas/cp
+grav         = 9.80665
+deg_to_rad   = math.pi/180.0
+kappa        = rdgas/cp
+p0           = 100000.0
+one_twelfth  = 1.0/12.0
+two_thirds   = 2.0/3.0
+
+missing_value = -9999.0 #9.99e20
+
+top_pres_for_forcing = 20000.0
+
+n_lam_halo_points = 3
+
+n_forcing_halo_points = 2 #number of points in each direction used to retrieve forcing (1 = 3x3, 2 = 5x5, 3 = 7x7); must be >=1
+
+#dist_method = 'euclidean'
+dist_method = 'haversine' #faster
+
+missing_variable_snow_layers = 3
+missing_variable_soil_layers = 4
+missing_variable_ice_layers = 2
+
+# Path to the directory containing processed case input files
+PROCESSED_CASE_DIR = '../../data/processed_case_input'
+
+# Path to the directory containing comparison data files
+COMPARISON_DATA_DIR = '../../data/comparison_data'
+
+# For developers: set logging level to DEBUG for additional output
+LOGLEVEL = logging.DEBUG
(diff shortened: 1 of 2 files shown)


dustinswales:
@grantfirl Thanks for creating this. I will fold my work on using the dynamic tendencies into this script.
One thing, it looks like the CI test for the replay is failing:
https://github.com/NCAR/ccpp-scm/actions/runs/8822685081/job/24221443644?pr=461
You need to update the environment file, environment-ufsreplay.yml, to have the new dependencies in your new script.

grantfirl:
> @grantfirl Thanks for creating this. I will fold my work on using the dynamic tendencies into this script. One thing, it looks like the CI test for the replay is failing: https://github.com/NCAR/ccpp-scm/actions/runs/8822685081/job/24221443644?pr=461 You need to update the environment file, environment-ufsreplay.yml, to have the new dependencies in your new script.

@dustinswales Sure, I can do this, although the purpose of this PR was to really just "check in" or show you guys this implementation. I think that the plan is still to compare the 1) dynamics tendencies vs. 2) total time tendency from history files (current method in the main repo) vs. 3) advective tendency from history files (what is in this new script) using the testing harness that you guys have created (or are creating). Now that we have methods 2 and 3 working in some form, should we try to run this through the test harness somehow? @scrasmussen @hertneky 

Like I said in the description, it's easy enough to fold this new script back into UFS_IC_generator.py if necessary for the test harness.

dustinswales:
> > @grantfirl Thanks for creating this. I will fold my work on using the dynamic tendencies into this script. One thing, it looks like the CI test for the replay is failing: https://github.com/NCAR/ccpp-scm/actions/runs/8822685081/job/24221443644?pr=461 You need to update the environment file, environment-ufsreplay.yml, to have the new dependencies in your new script.
> 
> @dustinswales Sure, I can do this, although the purpose of this PR was to really just "check in" or show you guys this implementation. I think that the plan is still to compare the 1) dynamics tendencies vs. 2) total time tendency from history files (current method in the main repo) vs. 3) advective tendency from history files (what is in this new script) using the testing harness that you guys have created (or are creating). Now that we have methods 2 and 3 working in some form, should we try to run this through the test harness somehow? @scrasmussen @hertneky
> 
> Like I said in the description, it's easy enough to fold this new script back into UFS_IC_generator.py if necessary for the test harness.

@grantfirl Gotcha. Then yes, let's run this through the harness. @hertneky @scrasmussen I will tag you in the google doc where you need to make changes.

grantfirl:
Combined into #473

PR #463, opened 2024-04-26
Add COMBLE case to CCPP SCM

bluefinweiwei:
- Forcing data was provide by COMBLE-MIP Project ([github repo](https://github.com/ARM-Development/comble-mip/blob/main/notebooks/forcing/COMBLE_INTERCOMPARISON_FORCING_V2.4.nc)). 
- @bluefinweiwei modified the forcing data and scm_input.F90 to work with CCPP SCM.

Forcing data:   ../data/processed_case_input/COMBLE_SCM_driver.nc
Case config file:  ../etc/case_config/COMBLE.nml
Modified:   scm_input.F90

attn: @hertneky @dustinswales @grantfirl @lulinxue


scm/src/scm_input.F90
```diff
@@ -1630,7 +1630,7 @@ subroutine get_case_init_DEPHY(scm_state, scm_input)
      call NetCDF_read_var(ncid, "wprvp_s",    .False., input_force_wprvp)
      call NetCDF_read_var(ncid, "wprtp_s",    .False., input_force_wprtp)
   else if (trim(input_surfaceForcingMoist) == 'surface_flux') then
-     call NetCDF_read_var(ncid, "hfls",       .False., input_force_sfc_lat_flx)
+     call NetCDF_read_var(ncid, "hfls",       .False., input_force_sfc_sens_flx)
```
grantfirl, line 1633: @bluefinweiwei What is the reason for this change? `hfls` is supposed to be the surface latent heat flux, not sensible.

bluefinweiwei: @grantfirl This change should be discarded. About this line, yes, you're right that it should be "lat" not "sens". Also, any change for this line is actually not necessary for COMBLE, since its surface forcing is surface temperature "ts" not heat fluxes. I am not sure why this was not updated and pushed, which might be a mistake by me. Sorry for the confusion!

grantfirl:
Combined into https://github.com/NCAR/ccpp-scm/pull/496

PR #465, opened 2024-05-01
Download climatological aerosol data and fix HR3 nml typo

grantfirl:
This PR adds a script in the `/contrib` directory to download the climatological aerosol data used when the first digit of iaer == 1. It also fixes a typo in the HR3 namelist name and adds iaer=1011 for testing HR3 (there may be other namelist changes forthcoming from @mzhangw )

Note, the script assumes that scm/data/physics_input_data directory already exists, so this should be used after the ./get_all_static_data.sh script.


dustinswales (approved):
This looks good to me. Thanks for doing this.
However, I wonder if we should just merge the thompson data script and this new aerosol dat fetching script into the get_all_ script?

grantfirl:
I thought about that too, but this script is adding like 12 GB of data and
I don't want to force everyone to download it if they don't need it. Same
for Thompson data.

mzhangw:
Thanks Grant. I tested with TWPICE with HR3 suite and both iaer = 1011 and 5111. It works well.
Screenshot 2024-05-02 at 10 38 43 AM

PR #474, opened 2024-05-16
Updates on HR3 suite for CCPP v7.0.0 release

mzhangw:
Namelist and tracer files updates on HR3 suite for CCPP v7.0.0 release.

PR #467, opened 2024-05-02
Add RRFS_v1 suite

grantfirl:
add SCM_RRFS_v1 SDF, nml, tracer file; update suite_info; update scm_type_defs to recognize new tracers

Source of SDF and tracer file: https://www.emc.ncep.noaa.gov/users/emc.campara/rrfs/namelist.html
Source of namelist: https://www.emc.ncep.noaa.gov/users/emc.campara/rrfs/namelists/input_namerica_detfcst.txt

PR #475, opened 2024-05-21
Update testing for release

dustinswales:
This PR contains changes/updates to the SCM RT CI for the CCPP V7 release. Includes the following:
- SCM RT suite list updated to supported V7 SDFs/NML/tracers configurations.
- Working CCPP v6 and v5 SDF/nml/tracer configurations are still run as part of the SCM RT suite, but listed as unsupported options (i.e. If/when they break, we don't fix em).
- Plotting script for SCM RT vs. Baseline comparison step. Plots of all SCM outputs for all supported cases and suites are created and uploaded to GitHub as an "artifact"

This PR contains changes from #467 and #474.


.github/workflows/ci_run_scm_rts.yml
```diff
@@ -150,6 +150,7 @@ jobs:
         cd ${SCM_ROOT}
         ./contrib/get_all_static_data.sh
         ./contrib/get_thompson_tables.sh
+        ./contrib/get_aerosol_climo.sh
```
grantfirl, line 153: This is going to slow down CI a lot. Any way to stash this data for future use?

test/cmp_rt2bl.py
```diff
@@ -25,6 +29,163 @@ def parse_args():
...
+                 "sfc_net_sw","sfc_dwn_lw", "u10m","v10m","hpbl","gflux",                \
+                 "qv_force_tend","T_force_tend","u_force_tend","v_force_tend","w_ls",    \
+                 "u_g","v_g","dT_dt_rad_forc","h_advec_thil","h_advec_qt",               \
+                 "v_advec_thil","v_advec_qt","T_s","lhf","shf","tprcp_inst",             \
+                 "tprcp_rate_inst","t2m","q2m","ustar","tsfc","tau_u","tau_v"]
+    # Open SCM datasets
+    SCM_BL = Dataset(file_BL)
+    SCM_RT = Dataset(file_RT)
+
+    plot_files = []
```
grantfirl, line 71: This is a huge value-add. Thanks, @dustinswales !

test/cmp_rt2bl.py
```diff
@@ -43,19 +204,39 @@ def main():
                 error_count = error_count + 1
             else:
                 print("Output for "+run["case"]+"_"+run["suite"]+ " is IDENTICAL to baseline")
+            # end if
+
+            # Create plots between RTs and baselines
+            plot_files = plot_results(file_rt,file_bl)
```
grantfirl, line 210: Can we move this to only create plots if results are different than baselines? Otherwise, it's kinda wasted effort.

test/cmp_rt2bl.py
grantfirl: Would this make sense to have it in its own module in case we want to use this outside of the RT context? E.g. generalize it a bit to just plot the output by itself (no comparison) using a GitHub label or starting the CI workflow manually?

grantfirl (approved):
Wonderful as-is, but there are a few questions to ponder.

.github/workflows/ci_run_scm_rts.yml
dustinswales: I was worried that this was going to bring this whole thing crashing down, but to my surprise it is faster to download this data on the GitHub servers than on Hera.[ It only took ~2min.](https://github.com/NCAR/ccpp-scm/actions/runs/9195025937/job/25289863152?pr=475), but that's more than the ~15sec before.
There may be slick ways to stash this, I just don't know any.

test/cmp_rt2bl.py
dustinswales: Good point. Will do

test/cmp_rt2bl.py
dustinswales: Good idea, I will see how much of this I can get in this afternoon.

dustinswales: done

test/cmp_rt2bl.py
dustinswales: done.

dustinswales:
@grantfirl I addressed your concerns. Now the script only creates plots when there are differences, and the comparison/plotting scripts are usable outside of the RT/CI context (I used them offline for faster debugging of the CI plots)
All good to merege!

PR #481, opened 2024-05-23
Add MOSAiC case to CCPP-SCM

hertneky:
This adds 2 cases for MOSAiC from Amy Solomon (amy.solomon@noaa.gov)

MOSAiC-AMPS: Arctic mixed-phase stratocumulus could (Oct 31 - Nov 5 2019)
MOSAiC-SS: Strongly stably stratified boundary layer (March 2-10 2020)

Data is in zip file: /scratch1/BMC/gmtb/Tracy.Hertneky/phys_tne/MOSAiC_SCM/MOSAiC_SCM.zip

Included are the:
- case configure nmls
- forcing file generation scripts
- scripts for running the SCM analysis and plotting

IMPROTANT
**Need to add the raw OBS files to the static data tarball with this PR - also processed case input which can be generated from the included scripts**


dustinswales:
@hertneky Thanks for adding these cases.
Can you sync your SCM branch with main?

grantfirl:
@hertneky Can you attach the raw data files that the scripts use or otherwise let me know where I can obtain them?

grantfirl:
Combined into https://github.com/NCAR/ccpp-scm/pull/496

PR #488, opened 2024-06-20
bugfix: CMake build missing mpi_f08 module paths

scrasmussen:
The fix 
- finds the MPI package and requires `mpi_f08` module. 
- include the MPI module paths, this fixes the following `use mpi_f08` error in `framework/src/ccpp_types.F90`

Current Error:
```
$ make
[  0%] Building Fortran object ccpp/framework/src/CMakeFiles/ccpp_framework.dir/ccpp_types.F90.o
/glade/work/soren/src/ccpp/upstream/scm-upstream/ccpp/framework/src/ccpp_types.F90:22:9:

   22 |     use mpi_f08, only: MPI_Comm
      |         1
Fatal Error: Cannot open module file 'mpi_f08.mod' for reading at (1): No such file or directory
compilation terminated.

```

### DIFF
--- a/scm/src/CMakeLists.txt
+++ b/scm/src/CMakeLists.txt
@@ -71,6 +71,10 @@ find_package(NetCDF REQUIRED COMPONENTS C Fortran)
 find_package(bacio REQUIRED)
 find_package(sp    REQUIRED)
 find_package(w3emc REQUIRED)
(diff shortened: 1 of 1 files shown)


grantfirl (approved):
Thanks, Soren. I was just running into this bug after I installed spack-stack on my local machine and was trying to compile.

PR #489, opened 2024-06-26
bugfix: NVidia Github Actions

scrasmussen:
Fixing Github runner nvidia MPI installation and caching. List of changes
- installing OpenMPI v4.1.6 manually, making MPI compilers default
- installing curl and zlibs with apt-get instead of manually
- bumping actions' versions
- quiet wget and tarball expansion
- small changes to caching and `PATH` and `LD_LIBRARY_PATH` variable setting. Separated from build steps

NOTE: nvidia-with-GPU build still fails

### DIFF
--- a/.github/workflows/ci_build_scm_ubuntu_22.04_nvidia.yml
+++ b/.github/workflows/ci_build_scm_ubuntu_22.04_nvidia.yml
@@ -8,6 +8,7 @@ jobs:
     # The type of runner that the job will run on
     runs-on: ubuntu-22.04
     strategy:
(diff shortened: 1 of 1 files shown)


grantfirl (approved):
This looks good to me @scrasmussen. It looks like it's failing at the point where #486 should fix it. @dustinswales Would you review this since you were working on this?

dustinswales (approved):
@scrasmussen 
This looks good to me. Thanks for fixing this.
(Hopefully this is the last of the non-containerized CI test we fix!)

PR #499, opened 2024-07-31
update case-gen scripts for LAM capability

hertneky:
Running the case generation on LAM output from UFS SRW does not currently work. These updates fix that.

Sample LAM data can be found on Hera: /scratch1/BMC/gmtb/Tracy.Hertneky/phys_tne/socrates/2018020412

Sample case over the S. Ocean was created with the following command:

./UFS_forcing_ensemble_generator.py -d /scratch1/BMC/gmtb/Tracy.Hertneky/phys_tne/socrates/2018020412 -n socrates -lons 133 -lats -51.5 -dt 75 -cres 768 -sdf SCM_GFS_v16 -sc -lam -fm 2 -vm 2

./run_scm.py --npz_type gfs --file scm_ufsens_socrates.py --timestep 75 -l 64 --n_itt_out 48 --n_itt_diag 48

Comparison of SCM to UFS column:
Diffs could be attributable to diff suite used (HRRR_gf for UFS vs GFS_v16 for SCM, since HRRR_gf was not working from case_gen input)
Screenshot 2024-07-31 at 3 04 52 PM

### DIFF
--- a/scm/etc/scripts/UFS_case_gen.py
+++ b/scm/etc/scripts/UFS_case_gen.py
@@ -459,12 +459,15 @@ def find_lon_lat_of_indices(indices, dir, tile, lam):
 ########################################################################################
 #
 ########################################################################################
-def find_loc_indices_UFS_history(loc, dir):
+def find_loc_indices_UFS_history(loc, dir, lam):
     """Find the nearest neighbor UFS history file grid point given a lon/lat pair"""
     #returns the indices of the nearest neighbor point in the given tile, the lon/lat of the nearest neighbor, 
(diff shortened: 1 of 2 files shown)


grantfirl (approved):
Looks good to me as long as you've tested on LAM output.

hertneky:
Tested using LAM output from Anders' SOCRATES case. Only tested fm=2, vm=2 (default). I can test other options if desired (just not "exact" mode).

dustinswales (approved):
These changes look identical to what I had done.

PR #500, opened 2024-08-08
NML changes for HRRR_gf and GFS_v17_p8_ugwpv1

grantfirl:
- contains namelist fixes for HRRR_gf to match what is used in the UFS RT hrrr_gf_debug
- contains namelist fixes for GFS_v17_p8_ugwpv1 to match what is used in the UFS RT control_p8_ugwpv1
-- differences: do_ca is set to F in the SCM, do_gwd_opt_psl is set to T in the SCM

Note: This PR is stacked on the dephy_conversion branch, so


ccpp/physics_namelists/input_GFS_v17_p8_ugwpv1.nml
dustinswales, line 1: FYI. When I added this nml, I followed the coupled simulation, https://github.com/ufs-community/ufs-weather-model/blob/develop/tests/rt.conf#L29, which is what I thought was the HR4 target, not the control_p8_ugwpv1 test.
Hence the differences here are not isolated to recent GWD updates.

dustinswales (approved):
@grantfirl Looks good to me.
Post-release we need to come up with a way to compare the SCM namelists with their UFS counterparts. 
This should be easy to do and help us keep our configurations in the SCM consistent with the UFS.

PR #504, opened 2024-08-15
Fix runtime_mult and Update remaining data retrieval scripts

grantfirl:
- Data retrieval scripts other than get_all_static_data.sh are now updated to grab data from v7.0.0-beta release assets
- @mkavulich reported that the `runtime_mult` run script option was no longer functional after conversion to DEPHY format
-- This was an oversight because the SCM was configured to calculate runtime from the begin/end date in the DEPHY files, ignoring whatever was in the case configuration namelist. In addition, runtime_mult is a run script-only function. Previously, it was simply used to modify the existing case configuration runtime, but for DEPHY cases, runtime wasn't even being provided in the case configuration namelists and would have been ignored anyway by the SCM.
-- These changes move the `runtime_mult` namelist option to be read in by the SCM if available.
-- The runtime listed in the case configuration namelist will be used (if given) even when the case uses the DEPHY format.
-- `runtime_mult` is used to modify the runtime in the SCM code, whether runtime comes from the case configuration namelist OR is calculated from the DEPHY file.

### DIFF
--- a/contrib/get_aerosol_climo.sh
+++ b/contrib/get_aerosol_climo.sh
@@ -20,7 +20,7 @@ data_files=("FV3_aeroclim1" "FV3_aeroclim2" "FV3_aeroclim3" "FV3_aeroclim_optics
 cd $BASEDIR/scm/data/physics_input_data/
 for file in "${data_files[@]}"; do
     echo "Retrieving $file.tar.gz"
-    wget https://github.com/NCAR/ccpp-scm/releases/download/v6.0.0/${file}.tar.gz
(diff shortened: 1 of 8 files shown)


dustinswales (approved):
@grantfirl Good catch! And thanks for reworking the runtime_multi option for DEPHY!

grantfirl:
@mkavulich One more change based on your feedback -- automatically grab the suites necessary for running RTs as the "default" build option rather than supported suites only.

mkavulich (approved):
@grantfirl thanks for getting this turned around so quickly. Runtimes seem to be behaving correctly, and the default suites are now identical to the RT suites.

hertneky (approved):
Glad this is fixed for DEPHY now!

PR #510, opened 2024-09-05
update plot_config file for tutorial

grantfirl:
The tutorial (specifically https://dtcenter.org/ccpp-scm-online-tutorial/exercise-1-run-scm-examine-results) needs these changes to work.

Without this, the plotting script will fail.


hertneky (changes requested):
Current changes look good, but I would go further to match other labels with the conversion factor.
L26
from:    conversion_factor = 1000.0, 1000.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0
to:   conversion_factor = 1000.0, 1000.0, 1.0, 86400., 86400., 86400., 86400., 86400.

For both profiles_mean_multi:
add:      conversion_factor = 86400.

The plots looked good with these additional changes.

PR #520, opened 2024-09-26
Move parse_tracers file from physics to host

dustinswales:
This PR moves the GFDL_parse_tracers file from the physics to the host (where it belongs)
Also, is a small bugfix in a plotting script.
@climbfuji


grantfirl:
@dustinswales @climbfuji Was doing something similar discussed for the fv3atm submodule? I'd like to not "get ahead" of the fv3atm if this hasn't been discussed, e.g. with Dusan. I agree that this is better, BTW, but modifying fv3atm isn't my call.

grantfirl:
@dustinswales I thought that the procedure for merging into NCAR/main was to make sure that the changes work for all hosts before merging, right? As-is, this breaks UWM without PRs there. I don't mind doing this for fv3atm/UWM because it's not difficult technically, as long as there is buy-in from fv3atm/UWM folks.

dustinswales:
> @dustinswales @climbfuji Was doing something similar discussed for the fv3atm submodule? I'd like to not "get ahead" of the fv3atm if this hasn't been discussed, e.g. with Dusan. I agree that this is better, BTW, but modifying fv3atm isn't my call.

We did not, but this is a no-brainer cleanup/housekeeping/error IMO, and something that I should've included when I did the repository organization. I
@DusanJovic-NOAA Sorry for not looping you in earlier, but are you okay with moving this host-dependent file from the ccpp-physics to the fv3atm?

DusanJovic-NOAA:
Are you suggesting adding function `get_tracer_index`, or similar, which is in this PR located in `module parse_tracers` to fv3atm repository?

climbfuji:
@DusanJovic-NOAA This is what I added for NEPTUNE in `ccpp/data` (similar directory structure as the UFS`) to accommodate our needs:
```
module parse_tracers

  integer, parameter :: NO_TRACER = -99

  public get_tracer_index, NO_TRACER

CONTAINS

  function get_tracer_index (tracer_names, gfdl_name, me, master, debug)

    character(len=32), intent(in) :: tracer_names(:)
    character(len=*),  intent(in) :: gfdl_name
    integer,           intent(in) :: me
    integer,           intent(in) :: master
    logical,           intent(in) :: debug
    !--- local variables
    character(len=32) :: name
    integer :: get_tracer_index
    integer :: i

    get_tracer_index = NO_TRACER

    do i=1, size(tracer_names)
       select case(trim(gfdl_name))
         ! Should not make it to here, since ntqv is hardcoded to 1
         !case '???'
         !  name = 'water_vapor_mixing_ratio'
         case ('o3mr')
           name = 'ozone_mixing_ratio'
         case ('liq_wat')
           name = 'cloud_water_mixing_ratio'
         case ('ice_wat')
           name = 'cloud_ice_mixing_ratio'
         case ('rainwat')
           name = 'rain_mixing_ratio'
         case ('snowwat')
           name = 'snow_mixing_ratio'
         case ('graupel')
           name = 'graupel_mixing_ratio'
         case ('cld_amt')
           name = 'cloud_fraction_updated_by_physics'
         case ('water_nc')
           name = 'number_concentration_of_cloud_droplets'
         case ('ice_nc')
           name = 'number_concentration_of_cloud_ic'
         case ('rain_nc')
           name = 'number_concentration_of_rain'
         case ('snow_nc')
           name = 'number_concentration_of_snow'
         case ('graupel_nc')
           name = 'number_concentration_of_graupel'
         case ('sgs_tke')
           name = 'turbulent_kinetic_energy'
         case ('liq_aero')
           name = 'number_concentration_of_water_friendly_aerosols'
         case ('ice_aero')
           name = 'number_concentration_of_ice_friendly_aerosols'
         case ('q_rimef')
           name = 'mass_weighted_rime_factor'
         case default
           name = trim(gfdl_name)
           if (debug .and. (me == master)) then
             print *,' PE ',me,' no translation found for tracer with GFDL name '//trim(gfdl_name)//' - use as is'
           endif
       end select
       if (trim(name) == trim(tracer_names(i))) then
           get_tracer_index = i
           exit
       endif
    enddo

    if (debug .and. (me == master)) then
      if (get_tracer_index == NO_TRACER) then
        print *,' PE ',me,' tracer with name '//trim(gfdl_name)//'='//trim(name)//' not found'
      else
        print *,' PE ',me,' tracer FOUND: '//trim(gfdl_name)//'='//trim(name)
      endif
    endif

    return

  end function get_tracer_index

end module parse_tracers
```

hertneky (approved):
Changes look good, but appears theres still ongoing conversation. Let me know if this should be merged or wait.

DusanJovic-NOAA:
I see several calls to the function named `get_tracer_index`:

https://github.com/search?q=repo%3ANOAA-EMC%2Ffv3atm%20get_tracer_index&type=code

Several of which are calls to the function with the same name from FMS and one in GFS_diagnostics.F90 and GFS_typedefs.F90 which looks like the one for this PR. Which can be confusing. If this version is only used from GFS_typedefs.F90, maybe we can just put it in that module.
BTW, as far as I can tell this function searchers for a string in a list of strings, and either return an index of the element if found, or -99 if the string is not in an array of string, nothing specific to tracers. IMO this can go in some strings utils module.

grantfirl:
> I see several calls to the function named `get_tracer_index`:
> 
> https://github.com/search?q=repo%3ANOAA-EMC%2Ffv3atm%20get_tracer_index&type=code
> 
> Several of which are calls to the function with the same name from FMS and one in GFS_diagnostics.F90 and GFS_typedefs.F90 which looks like the one for this PR. Which can be confusing. If this version is only used from GFS_typedefs.F90, maybe we can just put it in that module. BTW, as far as I can tell this function searchers for a string in a list of strings, and either return an index of the element if found, or -99 if the string is not in an array of string, nothing specific to tracers. IMO this can go in some strings utils module.

@dustinswales @climbfuji Now I remember why this PR was sitting for so long. The gfdl_parse_tracers isn't used in the SCM, I don't think. There is already a function in scm_type_defs for get_tracer_index(). I'm not sure that we need this in the SCM at all.

grantfirl:
> > @dustinswales I thought that the procedure for merging into NCAR/main was to make sure that the changes work for all hosts before merging, right? As-is, this breaks UWM without PRs there. I don't mind doing this for fv3atm/UWM because it's not difficult technically, as long as there is buy-in from fv3atm/UWM folks.
> 
> @grantfirl IMO We have a kinda wonky situation here that causes development to slow. The Root physics repository is used by two hosts, SCM and NEPTUNE, whereas the UWM uses its own fork. In this context the UWM is not a participating host, but instead we (the DTC) have a special code management agreement that puts an extra restriction on the Root that everything must pass in the Fork first.
> 
> I can see all kinds of problems with this arrangement. a) Why is a Fork, not the authoritative repository dictating the development workflow? b) How does someone contributing development to the Root run these tests? i) Unless the Root and all Forks are always synced, there is no easy way to do this, plus ii) outside developers don't have access to the NCAR forks that we use for testing. c) How is this going to work with multiple forks? (It wont)
> 
> I feel that this arrangement is unsustainable and needs to be revisited.

@dustinswales Sounds like a perfect subject for the Wednesday CCPP physics mgmt meeting!

grantfirl:
> > I see several calls to the function named `get_tracer_index`:
> > https://github.com/search?q=repo%3ANOAA-EMC%2Ffv3atm%20get_tracer_index&type=code
> > Several of which are calls to the function with the same name from FMS and one in GFS_diagnostics.F90 and GFS_typedefs.F90 which looks like the one for this PR. Which can be confusing. If this version is only used from GFS_typedefs.F90, maybe we can just put it in that module. BTW, as far as I can tell this function searchers for a string in a list of strings, and either return an index of the element if found, or -99 if the string is not in an array of string, nothing specific to tracers. IMO this can go in some strings utils module.
> 
> @dustinswales @climbfuji Now I remember why this PR was sitting for so long. The gfdl_parse_tracers isn't used in the SCM, I don't think. There is already a function in scm_type_defs for get_tracer_index(). I'm not sure that we need this in the SCM at all.

OK, the situation is little messier than this. GFS_typedefs DOES use the get_tracer_index() routine from the parse_tracers module but scm_type_defs has its own. Let me think how best to clean this up.

dustinswales:
> > I see several calls to the function named `get_tracer_index`:
> > https://github.com/search?q=repo%3ANOAA-EMC%2Ffv3atm%20get_tracer_index&type=code
> > Several of which are calls to the function with the same name from FMS and one in GFS_diagnostics.F90 and GFS_typedefs.F90 which looks like the one for this PR. Which can be confusing. If this version is only used from GFS_typedefs.F90, maybe we can just put it in that module. BTW, as far as I can tell this function searchers for a string in a list of strings, and either return an index of the element if found, or -99 if the string is not in an array of string, nothing specific to tracers. IMO this can go in some strings utils module.
> 
> @dustinswales @climbfuji Now I remember why this PR was sitting for so long. The gfdl_parse_tracers isn't used in the SCM, I don't think. There is already a function in scm_type_defs for get_tracer_index(). I'm not sure that we need this in the SCM at all.

@grantfirl That's correct, we don't need any host changes for the SCM, but any model that uses the Root needs to be tested with this physics change.

dustinswales:
> > > I see several calls to the function named `get_tracer_index`:
> > > https://github.com/search?q=repo%3ANOAA-EMC%2Ffv3atm%20get_tracer_index&type=code
> > > Several of which are calls to the function with the same name from FMS and one in GFS_diagnostics.F90 and GFS_typedefs.F90 which looks like the one for this PR. Which can be confusing. If this version is only used from GFS_typedefs.F90, maybe we can just put it in that module. BTW, as far as I can tell this function searchers for a string in a list of strings, and either return an index of the element if found, or -99 if the string is not in an array of string, nothing specific to tracers. IMO this can go in some strings utils module.
> > 
> > 
> > @dustinswales @climbfuji Now I remember why this PR was sitting for so long. The gfdl_parse_tracers isn't used in the SCM, I don't think. There is already a function in scm_type_defs for get_tracer_index(). I'm not sure that we need this in the SCM at all.
> 
> OK, the situation is little messier than this. GFS_typedefs DOES use the get_tracer_index() routine from the parse_tracers module but scm_type_defs has its own. Let me think how best to clean this up.



> > > I see several calls to the function named `get_tracer_index`:
> > > https://github.com/search?q=repo%3ANOAA-EMC%2Ffv3atm%20get_tracer_index&type=code
> > > Several of which are calls to the function with the same name from FMS and one in GFS_diagnostics.F90 and GFS_typedefs.F90 which looks like the one for this PR. Which can be confusing. If this version is only used from GFS_typedefs.F90, maybe we can just put it in that module. BTW, as far as I can tell this function searchers for a string in a list of strings, and either return an index of the element if found, or -99 if the string is not in an array of string, nothing specific to tracers. IMO this can go in some strings utils module.
> > 
> > 
> > @dustinswales @climbfuji Now I remember why this PR was sitting for so long. The gfdl_parse_tracers isn't used in the SCM, I don't think. There is already a function in scm_type_defs for get_tracer_index(). I'm not sure that we need this in the SCM at all.
> 
> OK, the situation is little messier than this. GFS_typedefs DOES use the get_tracer_index() routine from the parse_tracers module but scm_type_defs has its own. Let me think how best to clean this up.

Can't we just stick `get_tracer_index` into GFS_typedefs?

climbfuji:
> I see several calls to the function named `get_tracer_index`:
> 
> https://github.com/search?q=repo%3ANOAA-EMC%2Ffv3atm%20get_tracer_index&type=code
> 
> Several of which are calls to the function with the same name from FMS and one in GFS_diagnostics.F90 and GFS_typedefs.F90 which looks like the one for this PR. Which can be confusing. If this version is only used from GFS_typedefs.F90, maybe we can just put it in that module. BTW, as far as I can tell this function searchers for a string in a list of strings, and either return an index of the element if found, or -99 if the string is not in an array of string, nothing specific to tracers. IMO this can go in some strings utils module.

Whatever solution works best for the UFS - the main point is to move it out of ccpp-physics.

grantfirl:
@dustinswales @climbfuji @DusanJovic-NOAA We're definitely right to remove GFDL_parse_tracers.F90 from ccpp-physics. For fv3atm, we should just use the one from FMS. The one used by GFS_typedefs from the physics was totally redundant. For the SCM, we can use the version already included in scm_type_defs.F90 since we don't use FMS. I'm going to add PRs into fv3atm and SCM for this and close this one. @dustinswales I'll add the bugfix in cmp_scmout to the new one.

dustinswales:
@grantfirl Sounds good to me. Thanks!

grantfirl:
Replaced by https://github.com/NCAR/ccpp-scm/pull/569. The fv3atm PR is coming shortly.

PR #522, opened 2024-09-26
Capgen in scm

dustinswales:


ccpp/suites/suite_SCM_GFS_v17_p8_ugwpv1.xml
dustinswales, line 1: @climbfuji Here's the snippet that I added to the SDF which you could drop into the rest of them.

ccpp/suites/suite_SCM_GFS_v17_p8_ugwpv1.xml
```diff
@@ -59,6 +59,12 @@
       <scheme>ugwpv1_gsldrag</scheme>
       <scheme>ugwpv1_gsldrag_post</scheme>
       <scheme>GFS_GWD_generic_post</scheme>
+    </subcycle>
+  </group>
+  <!-- End process-split section -->
+  <!-- Begin time-split section -->
+  <group name="physics_ts">
+    <subcycle loop="1">
       <scheme>GFS_suite_stateout_update</scheme>
```
climbfuji, line 68: Should the GFS_suite_stateout_update be part of the process_split group? Seems more logical to me that the process split group at the end updates the state from all the processes (this is also what the framework will do automatically in the future)?

dustinswales: Yeah, that makes more sense.

PR #526, opened 2024-10-10
Contrib data script's wget verbosity

scrasmussen:
## Issue
The Github Actions logs are being filled up with output from the `wget` download status bar.

## Description of Changes
Added ability to turn on verbose wget output. Wget output defaults to `-nv` so that the Github Actions output won't be filled with the wget download status. This will make debugging failed testcases easier.

Since I was adding command line arguments I also added `--help` option. 

## Question for Reviewers
Some of the script have `set -ex` and some have it commented out. Should we default to the same behavior for all the scripts?

## Tests Conducted: 
Tested scripts by hand.

### DIFF
--- a/contrib/get_aerosol_climo.sh
+++ b/contrib/get_aerosol_climo.sh
@@ -1,5 +1,35 @@
 #!/bin/bash
 
+# Function to display help message
+print_help() {
+    echo "get_aerosol_climo.sh: contrib/get_aerosol_climo.sh [-v,--verbose]"
+    echo "    Script for downloading/extracting the GOCART climatological aerosol data."
+    echo ""
+    echo "Options:"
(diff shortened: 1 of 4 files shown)


dustinswales (approved):

dustinswales (commented):
@scrasmussen Thanks for adding the functionality to pass in the verbosity option. 
One request, could we add this to the CI scripts as part of this PR? [here](https://github.com/NCAR/ccpp-scm/blob/main/.github/workflows/ci_run_scm_rts.yml#L158) and [here](https://github.com/NCAR/ccpp-scm/blob/main/.github/workflows/ci_run_scm_DEPHY.yml#L152)?

dustinswales:
@scrasmussen Ignore my last comment. (I forgot that we want it to be "-nv" by default)

PR #529, opened 2024-10-17
UFS-dev PR#183

grantfirl:
Changes similar to https://github.com/NOAA-EMC/fv3atm/pull/798

Additional changes:
- Make the interstitial DDT in the SCM physics DDT an array of (n_threads) size so that GFS_Debug can work (this is how UFS does it)
-- downstream changes for the above change

RT notes: This PR doesn't change RT results except for the addition of a new test that exercise the GFS_debug.F90 schemes. Note that the `GFS_abort` scheme in GFS_debug.F90 is untested because it causes a run failure (on purpose) and CI flags it as a failure, which we don't necessarily want.

### DIFF
--- a/.github/workflows/ci_run_scm_rts.yml
+++ b/.github/workflows/ci_run_scm_rts.yml
@@ -22,8 +22,8 @@ jobs:
       sp_ROOT:    /home/runner/NCEPLIBS-sp
       w3emc_ROOT: /home/runner/myw3emc
       SCM_ROOT:   /home/runner/work/ccpp-scm/ccpp-scm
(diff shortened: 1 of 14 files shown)


dustinswales (approved):

grantfirl:
@dustinswales @mkavulich @scrasmussen RTs run as expected now. There were a couple of failures related to the original PR -- one was that I failed to compile the new suite(s) in the RT CI workflow and the second was that, for testing purposes, I include the `GFS_abort` scheme from GFS_debug.F90 in the new suites, but this sets a non-zero error flag upon run termination and the CI flags it as a failure. All of that is expected, but we don't want to have CI fail from henceforth, so I just removed the `GFS_abort` scheme. All is good at this point, so I'll go ahead and merge after the ccpp-physics one.

PR #533, opened 2024-10-18
UFS-dev PR#226

grantfirl:
This is only a ccpp-physics submodule update for https://github.com/NCAR/ccpp-physics/pull/1097

RT baseline changes are consistent with changes to ccpp physics changes.

PR #546, opened 2024-12-05
Update ccpp/framework hash

grantfirl:
This is only updating the ccpp/framework hash in coordination with https://github.com/NCAR/ccpp-framework/pull/615


grantfirl:
RTs are differing from baselines, but (I think) only because baseline artifacts were not updated after #533 and later PRs. I'm going to run them manually on my Mac since I can't update the baselines on the webserver.

Issue #531, opened 2024-10-18
casegen error when supplying ij indices


hertneky:
When running casegen using LAM data, @egrell discovered that the UFS_case_gen.py script fails if providing ij indices. The same failure also exists when using global RTs. 

The issue is **_main()_** calling **_find_loc_indices_UFS_history()_**, which only has logic to find the indices nearest to the provided lon/lat point, hence why latitude/longitude are not subscriptable, since we are only providing ij. On the other hand, **_find_loc_indices_UFS_IC_** does have logic to handle either provided lon/lat or indicies. Can this similar logic be added to the problematic subroutine?

Note, the ensemble wrapper script does not even have an option for providing ij.

Example command:
`./UFS_case_gen.py  -ij 142 69 -d 20210322060000 -i /scratch1/BMC/gmtb/Tracy.Hertneky/ccpp/case_gen/ufs/control_c48_intel_1h/INPUT -g /scratch1/BMC/gmtb/Tracy.Hertneky/ccpp/case_gen/ufs/control_c48_intel_1h/INPUT -f /scratch1/BMC/gmtb/Tracy.Hertneky/ccpp/case_gen/ufs/control_c48_intel_1h -n test_ij_rt -t 5`

Error:
```
Traceback (most recent call last):
  File "./UFS_case_gen.py", line 3773, in <module>
    main()
  File "./UFS_case_gen.py", line 3688, in main
    (hist_i, hist_j, hist_lon, hist_lat, hist_dist_min, angle_to_hist_point, neighbors, dx, dy) = find_loc_indices_UFS_history(location, forcing_dir, lam)
  File "./UFS_case_gen.py", line 506, in find_loc_indices_UFS_history
    dist[i,j] = haversine_distance(latitude[i,j],longitude[i,j],loc[1],loc[0])
TypeError: 'NoneType' object is not subscriptable
```

hertneky:
@egrell Was advised to use --location and --use_nearest to get the specified column for now.

grantfirl:
> When running casegen using LAM data, @egrell discovered that the UFS_case_gen.py script fails if providing ij indices. The same failure also exists when using global RTs.
> 
[...]

PR #549, opened 2025-01-07
ij index fix for issue #531

grantfirl:
- reinstate functionality to give UFS_case_gen.py an i,j index pair along with tile number to generate a case
- add ability to provide UFS_forcing_ensemble_generator.py with a list of i,j indices to generate multiple cases using the i,j,tile paradigm
- update documentation to reflect reinstated capability

fixes https://github.com/NCAR/ccpp-scm/issues/531


grantfirl:
> @grantfirl I tried the -ij option to extract the forcing for my selected points. While the code ran successfully, it didn't give me the locations that I wanted. I apologize if I perhaps wasn't clear in describing my use of the option. I would like to choose the points based on what I can see in my model (SRW) output, ie via ncview, which makes it easy to identify the points. The problem, I think, is that the case_gen script is first finding the points in the $grid_dir/grid.tile{0}.halo{1}.nc file, which is on the supergrid, not on the model grid. I see that the code includes an accounting for the halo (I have n_lam_halo_points = 3, the default), and extraction of every other point, but it still isn't coming out quite right.
> 
> I wonder if either: a) the code could be re-organized to call find_loc_indices_UFS_history first, and set the location from the history files rather than the input files, (I think this was done previously?) or b) Do I have some settings wrong regarding the INPUT grid? My model output has dimensions: grid_xt = 1000 ; grid_yt = 780 ; but the INPUT/_grid.tile7.halo3.nc file has dimensions 2060 x 1612. These settings are all defaults. I don't quite see how subtracting the halo_2 and taking every other point gets me to the matching grid.

@egrell Do you have any of the data saved somewhere on Hera that I could copy and test with to try to understand better what is going wrong? 

From your description, it sounds like what was intended for the -ij option just isn't what you're really looking for. This option is supposed to correspond to the native grid, not the output grid. We could probably add yet another option like "-ij_history" or something to do what you need.

Nevertheless, I'd like to use some of your data to see if the algorithm is doing what was intended. It's certainly possible that something has changed in the SRW grid configuration to make the code wrong (or maybe it was never right - always possible).

grantfirl:
@egrell @hertneky As far as I can tell, the code is still doing what is intended. Providing the i and j indices correspond to the FV3 native grid, and this appears to still be correct. The script is looking in CXXXX_grid.tile7.halo3.nc to read in the supergrid, including halo points. Halo points are subtracted and every other point is kept, giving one the native grid. This all appears to be correct, as long as one sets n_halo_points in the script to correspond to the available CXXXX_grid.tile7.haloY.nc file correctly.

Note that the native grid need not have much of a relationship to the output file grid (see https://ufs-weather-model.readthedocs.io/en/develop/InputsOutputs.html#output-files). The write component takes the output on the native grid and writes the output files how they are specified in `model_configure`. For the given data from @egrell, the write component produced files of size (1000, 780) with the following parameters (these are global attributes in the dynf/phyfXXX.nc files):

                :cen_lat = 39.5f ;
                :cen_lon = -83.f ;
                :dx = 3000.f ;
                :dy = 3000.f ;
                :grid = "lambert_conformal" ;
                :grid_id = 1LL ;
                :hydrostatic = "non-hydrostatic" ;
                :lat1 = 27.5f ;
                :lon1 = -98.f ;
                :ncnsto = 10 ;
                :nx = 1000 ;
                :ny = 780 ;
                :source = "FV3GFS" ;
                :stdlat1 = 39.5f ;
                :stdlat2 = 39.5f ;

This grid is centered on the same point as the native grid, but uses a different projection. It also removes 20 points in the y direction and 24 points in the x direction, probably to not output data with boundary effects(?). Note that it is possible to configure the write component to write out the FV3 native grid, then the i,j indices should correspond to one-to-one to the native grid calculated from CXXXX_grid.tile7.haloY.nc, but this is not the default.

If what is desired is to take indices from the files written from the write component (currently given names of dynf/phyfXXX.nc), translate those indices into lat/lon, then grab the closest point on the native grid, that is also possible, but it is basically the reverse of what is currently being done, and this doesn't really constitute a bug fix, but an added feature.

@hertneky I've now tested the UFS_forcing_ensemble_generator.py addition and it works on Hera after a couple of bug fixes. This should be ready to review/approve/merge for the bug fix release.

grantfirl:
> Note: Still need to test UFS_forcing_ensemble_generator.py functionality on Hera.

This has been done. It works fine as long as #563 is used.

hertneky (approved):
Code changes look good. Tests provided expected results, checking that the correct point was extracted from the native grid. SCM from both produced data files ran fine.

Tested both UFS_case_gen.py:
`./UFS_case_gen.py -ij 24 24 -t 4 -i /scratch1/BMC/gmtb/Dustin.Swales/SCM/replay/rt_3252959/control_c48_intel/INPUT/ -g /scratch1/BMC/gmtb/Dustin.Swales/SCM/replay/rt_3252959/control_c48_intel/INPUT/ -f /scratch1/BMC/gmtb/Dustin.Swales/SCM/replay/rt_3252959/control_c48_intel -n control_c48_intel_C_test_n000 -sc -fm 2 -vm 2`

And ./UFS_forcing_ensemble_generator.py:
`./UFS_forcing_ensemble_generator.py -d /scratch1/BMC/gmtb/Dustin.Swales/SCM/replay/rt_3252959/control_c48_intel -sc --C_RES 48 -dt 1200 -n control_c48_intel_C_test -sdf SCM_GFS_v17_p8 -tile 4 -is 24 -js 24`

grantfirl:
@hertneky @scrasmussen Are we ready to merge this?

hertneky:
@grantfirl I'm good with merging.

scrasmussen (approved):
lgtm 👍  the code looks good, I didn't run it but Tracy has tested it [here](https://github.com/NCAR/ccpp-scm/pull/549#pullrequestreview-2648750969) so I think we are good to merge.

PR #562, opened 2025-02-20
warning fixes: Github Actions Updates

scrasmussen:
Changes to fix warnings from Github CI runs. 

- Change all to latest Ubuntu runner option, 20.04 is being deprecated.

- Simplifying test names and renaming files for move to latest Ubuntu runner

- Testing with Spack Python versions 3.11.7 and latest

- Updating NVIDIA HPC SDK release, only building and not using Conda to reduce space

- Adding more unique cache names

- Fixes warning if plots aren't created

- ftp site was taken down, moving location of replay

### DIFF
--- a/.github/workflows/ci_build_scm_ubuntu.yml
+++ b/.github/workflows/ci_build_scm_ubuntu.yml
@@ -1,4 +1,4 @@
-name: CI test to build the CCPP-SCM on ubuntu v22.04
+name: build the CCPP-SCM on latest Ubuntu runner
 
 on: [pull_request,workflow_dispatch]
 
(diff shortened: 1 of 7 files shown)


.github/workflows/ci_build_scm_ubuntu_22.04_nvidia.yml
dustinswales, line 1: Since we know this test will always fail, should we turn if off? Or at least to only run on workflow_dispatch?
(I really hate the red x!!!)

.github/workflows/ci_build_scm_ubuntu_22.04_nvidia.yml
```diff
@@ -1,19 +1,19 @@
-name: CI test to build the CCPP-SCM on ubuntu v22.04
+name: build the CCPP-SCM on ubuntu v22.04
```
dustinswales, line 1: I thought v20.04 is being sunset, not v22.04?
Anyways, we're using the latest version now. Remove version number from name

dustinswales (commented):
Thanks for making these changes. I had a few minor suggestions.

.github/workflows/ci_build_scm_ubuntu_22.04_nvidia.yml
scrasmussen: Agreed! Will turn it off

.github/workflows/ci_build_scm_ubuntu_22.04_nvidia.yml
dustinswales: Do it!

grantfirl (approved):
Looks good to me. Lots of good work here!

PR #565, opened 2025-03-10
Bug fix in physics/GWD/ugwp_driver_v0.F for NEPTUNE

grantfirl:
See https://github.com/NCAR/ccpp-physics/pull/1104 for the description.

This change should not affect the SCM at all since ugwpv0 is not called in any suite.


grantfirl:
RTs were identical the first time they ran.  No need to update baselines on mohawk.

hertneky (approved):

PR #566, opened 2025-03-11
UFS dev PR#232/256

hertneky:
Changes combined from PRs 232 and 256

Associated fv3atm PRs:
NOAA-EMC/fv3atm [#993](https://github.com/NOAA-EMC/fv3atm/pull/933)
NOAA-EMC/fv3atm [#885](https://github.com/NOAA-EMC/fv3atm/pull/885)

PR #569, opened 2025-03-18
Updated version of "Move parse_tracers file from physics to host"

grantfirl:
Original description/work from @dustinswales:

This PR moves the GFDL_parse_tracers file from the physics to the host (where it belongs)
Also, is a small bugfix in a plotting script.

@grantfirl update:

- remove parse_tracers module entirely since we can use a local version of get_tracer_index() already in SCM code
- move get_tracer_index() function from scm_type_defs to GFS_typedefs to avoid circular dependency

### DIFF
--- a/scm/src/GFS_typedefs.F90
+++ b/scm/src/GFS_typedefs.F90
@@ -29,7 +29,8 @@ module GFS_typedefs
    integer, parameter :: dfi_radar_max_intervals = 4 !< Number of radar-derived temperature tendency and/or convection suppression intervals. Do not change.
 
    real(kind=kind_phys), parameter :: limit_unspecified = 1e12 !< special constant for "namelist value was not provided" in radar-derived temperature tendency limit range
-
+   
+   integer, parameter :: no_tracer = -99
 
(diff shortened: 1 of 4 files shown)


grantfirl:
@mkavulich @scrasmussen @hertneky @dustinswales This introduces another "delta" between the fv3atm and SCM version of GFS_typedefs. The fv3atm version (PR to come) will use get_tracer_index() from FMS (which is not used in SCM) like the rest of fv3atm instead of the previously included version from GFDL_parse_tracers in ccpp-physics.

grantfirl:
Replaces https://github.com/NCAR/ccpp-scm/pull/520

scrasmussen (approved):
Looked over changes, `get_tracer_index` is an equivalent and slightly cleaner version to the one removed from the `ccpp-physics` repo. Looks good and compiles!

grantfirl:
The same RTs from https://github.com/NCAR/ccpp-scm/pull/566 are expected to change results here since new baselines haven't been uploaded to mohawk yet.

Issue #519, opened 2024-09-20
DEPHY format with specified fluxes


wangevine:
I am trying to run a new case using the DEPHY format with specified fluxes.  There are two issues:
1. If I specify surface_type ocean, there is no flux and the fluxes in the output file are identically zero.
2. If I specify surface_type land, the fluxes in the output file are correct, but the temperature in the model decreases rapidly.  It may be that the model is using large negative fluxes (maybe the missing value?).

All advection and nudging terms are turned off.  Both issues occur with either the HRRR or GFS_v17_p8 suite.

Everything is on Hera under /scratch2/BMC/calnexfc/angevine/CCPP_SCM_v7.0
The case is called samens_015.  Several versions of the input file are in data/processed_case_input/dephy_force_samens*.  The one that I used most recently was copied to samens_015_SCM_driver.nc.

grantfirl:
@wangevine I'll take a look to see if I can see where the problem lies.

grantfirl:
@wangevine I've had some success debugging the issues. I used the samens_015_SCM_driver.nc from that location on Hera that you shared. Most of the problems that I was running into were due to missing data in the file.

1. The values for nudging_X in the global attributes were set to nonzero values, yet the associated X_nud data variables were missing.
2. The `surface_forcing_wind` needs to either be `z0` or `ustar`. `none` isn't implemented and could lead to problems.
3. There is a bug/missing feature with respect to missing variables for the case input data file. The SCM code is written to expect that missing variables are set to -9999 rather than the default netCDF ones. I ran into an issue where pa_forc was missing in the netCDF file, but the SCM code didn't recognize it since the missing values were treated as some huge value by default, but it was expecting -9999 to be able to handle it. This can be fixed in the SCM code and I'll add an issue to do that. In the meantime, if you set whatever missing values you have to -9999, that might get around some problems.
4. Finally, at least for the suite that I was using, the PBL scheme divides by the wind_speed. Since both u and v were set to 0, this was causing NaNs. I've discovered that for non-prescribed surface flux suites, the wind speed is augmented by a convective gustiness factor and/or given a minimum value of 1, so the PBL scheme never divides by zero. For the prescribed surface flux suites, the calculation of wind didn't have that "safety" feature, but I should add this to the scm_sfc_flux_spec scheme to prevent this issue. In the meantime, you should be able to get past this error by setting u and v to some small non-zero value.

[...]

PR #570, opened 2025-03-19
enhancement/bugfix: use NetCDF missing values instead of -9999

scrasmussen:
DESCRIPTION OF CHANGES:
 - Using NetCDF variables for missing_values instead of `-9999.0` and `-9999`
 - Fixing/adding generic interfaces for single and double precision. Because the generic interface allow a single subroutine name to be used to call using variables of single or double precision, `missing_value_sp` and `missing_value_dp` variables need to be used in the subroutine definitions.
 - Added generic interface definitions for `NetCDF_read_att_dp`, `NetCDF_put_var_1d_sp`, `NetCDF_put_var_2d_sp`, `NetCDF_put_var_1d_dp`, `NetCDF_put_var_2d_dp`, `check_missing_real_sp_0d`, `check_missing_real_sp_1d`, `conditionally_set_real_sp_var_0d`, `conditionally_set_real_sp_var_1d`

ISSUES: Related to issues brought up in Issue #519 

TESTS: compiles both single and double precision build setup

### DIFF
--- a/.github/workflows/ci_build_scm_ubuntu_nvidia.yml
+++ b/.github/workflows/ci_build_scm_ubuntu_nvidia.yml
@@ -1,11 +1,18 @@
 name: build the CCPP-SCM with Nvidia
 
-on: [pull_request,workflow_dispatch]
+on:
+  workflow_run:
+    workflows: ["build and run SCM regression tests"]
+    types:
+      - completed
 
 jobs:
 
   build_scm:
-    # The type of runner that the job will run on
+    # trigger only after the SCM regression tests have completed from a PR or workflow event
+    if: |
+      github.event.workflow_run.event == 'pull_request' ||
+      github.event.workflow_run.event == 'workflow_dispatch'
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
--- a/.github/workflows/ci_test_docker.yml
+++ b/.github/workflows/ci_test_docker.yml
@@ -1,6 +1,10 @@
 name: build test and push docker
 
-on: [pull_request,workflow_dispatch]
+on:
+  # not listing pull_request closed, since it would run when merged
+  pull_request:
+    types: [opened, synchronize, reopened]
+  workflow_dispatch:
 
 env:
   TEST_TAG: dtcenter/ccpp-scm:test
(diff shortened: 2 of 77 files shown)


dustinswales (approved):
Looks good to me. 
The precision police abide :)

scrasmussen:
> Is there issue with checking if var == missing_value, as I see in other places in the code? I understand that this method may not be reliable due to precision errors of floating point numbers, so probably not!

@hertneky 
If the comparison uses variables of a different data type there might be an issue but this can be worked around with generic interfaces.

This [commit](https://github.com/NCAR/ccpp-scm/pull/570/commits/ff31f8e490ffb80f7e3884f05caad320f7d6946b) adds generic procedures to report true/false if a variable is a missing value. The generic interfaces mean that it can choose the same datatype for MISSING_VALUE as the input value. Using a `is_missing_value` procedure also helps with readability.

grantfirl:
@scrasmussen I think that https://github.com/scrasmussen/ccpp-scm/pull/8 should fix whatever issue was showing up in the RTs. I ran them on my local machine and they ran fine.

grantfirl:
@scrasmussen CI RTs are still failing. Hmm. I can't reproduce the failure on my machine, which is odd. I'm re-running the debug mode RTs only to see if it's only the single precision tests failing (which kills the associated tests that take longer).

grantfirl:
@scrasmussen Still working on a fix for this. There are many places that had been assuming -999 that still exist in the code that I'm fixing. There will be another PR into your branch shortly.

grantfirl:
@scrasmussen @hertneky Please see https://github.com/scrasmussen/ccpp-scm/pull/9. It updates this PR branch to the latest main and has many fixes for surface variable initialization that no longer rely on checking for -999 or <0.

grantfirl:
@scrasmussen @hertneky Using nccmp -d on manually-run RT output shows that the differences are only in the missing data values, as hypothesized. I didn't check all files, but the ones that I did showed this.

grantfirl:
> @scrasmussen @hertneky Using nccmp -d on manually-run RT output shows that the differences are only in the missing data values, as hypothesized. I didn't check all files, but the ones that I did showed this.

Some tests don't show any differences with `nccmp -d -f` (the data is identical) but do show differences with `nccmp -m -f` (metadata is different - specifically missing value attributes). This is as-expected, and this can be merged.

grantfirl:
@hertneky @scrasmussen @mkavulich Please update the baselines on mohawk with artifacts from https://github.com/NCAR/ccpp-scm/actions/runs/14599498366

scrasmussen:
> @hertneky @scrasmussen @mkavulich Please update the baselines on mohawk with artifacts from https://github.com/NCAR/ccpp-scm/actions/runs/14599498366

Updated the baselines, thanks for your work on this @grantfirl!

PR #576, opened 2025-03-27
NSST initialization changes

grantfirl:
Following discussion with @hertneky, this changes the following:

- Change the initial value of the ocean mixed layer depth (`xz`) to 20.0 m to match UFS
- Change `ifd` or the flag to turn on NSST to 1.0 if the surface is ocean; this only has an effect if prescribed surface fluxes are off and NSST is in the SDF and nstf_name(1) > 0.
-- if `ifd` starts as zero, NSST code can eventually set it to 1 if the model time is after 6 AM local time and ifd starts as 0, but during some cases, it was noticed that `ifd` wouldn't change to 1 (and therefore activate NSST) until the first "sunrise" after the model initializes (e.g. if a case started at noon, `ifd` wouldn't set itself to 1 until the next morning, even though NSST should be active immediately)
 - Set the nstf_name(2) value to 1 (NSST spin-up) in every SCM physics namelist to be consistent; the only chance for the SCM to not need spin-up is for a UFS-based case that has all of the NSST variables ready to be read in; **NOTE: for UFS cases needing to use NSST and when NSST variables are included in the case data file, the nstf_name(2) should be set to 0** (I don't think that this is documented anywhere. We can also probably make sure this happens programatically in the UFS_case_gen scripts. Added https://github.com/NCAR/ccpp-scm/issues/577 to address in the future).

New functionality needed for @bluefinweiwei:

- Add `do_sst_initialize_only` namelist variable (default value of F) to the case configuration namelist; if set to T, scm_state%T_surf (which gets pointed to by the `sea_surface_temperature` variable) only gets initialized from the case data file SST value. The default/original behavior is that when the `surface_forcing_temp` variable in the DEPHY-formatted case data file is set to `ts` (or any of the other options too, for that matter), the scm_state%T_surf gets updated from the case forcing file every timestep and physics has very little "authority" over this variable -- it may get changed during the physics call, but it will get reset when forcing occurs. The new behavior allows physics (including NSST) to change sea_surface_temperature how it wants to.

### DIFF
--- a/scm/src/scm_forcing.F90
+++ b/scm/src/scm_forcing.F90
@@ -326,7 +326,7 @@ subroutine interpolate_forcing(scm_input, scm_state, in_spinup)
           end do
         end if
 
-        if (scm_state%surface_thermo_control == 0 .or. scm_state%surface_thermo_control == 1 .or. scm_state%surface_thermo_control == 2 .and. .not. scm_state%do_sst_initialize_only) then
+        if (scm_state%surface_thermo_control == 0 .or. scm_state%surface_thermo_control == 1 .or. scm_state%surface_thermo_control == 2 .and. (.not. scm_state%do_sst_initialize_only)) then
           !skin temperature is needed if surface fluxes are specified (for calculating bulk Richardson number in the specified surface flux scheme) and for simple ocean scheme
           do i=1, scm_state%n_cols
             scm_state%T_surf(i) = scm_input%input_T_surf(scm_input%input_ntimes)
--- a/scm/src/scm_forcing.F90
+++ b/scm/src/scm_forcing.F90
@@ -144,7 +144,9 @@ subroutine interpolate_forcing(scm_input, scm_state, in_spinup)
 
           !>  - Set the surface parameters to the last available data.
           scm_state%pres_surf(i) = scm_input%input_pres_surf(scm_input%input_ntimes)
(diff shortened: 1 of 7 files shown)


scm/src/scm_forcing.F90
```diff
@@ -324,7 +326,7 @@ subroutine interpolate_forcing(scm_input, scm_state, in_spinup)
           end do
         end if
 
-        if (scm_state%surface_thermo_control == 0 .or. scm_state%surface_thermo_control == 1 .or. scm_state%surface_thermo_control == 2) then
+        if (scm_state%surface_thermo_control == 0 .or. scm_state%surface_thermo_control == 1 .or. scm_state%surface_thermo_control == 2 .and. .not. scm_state%do_sst_initialize_only) then
```
scrasmussen, line 329: Because of operator precedence this is an equivalent statement to `if (scm_state%surface_thermo_control == 0 .or. scm_state%surface_thermo_control == 1 .or. (scm_state%surface_thermo_control == 2 .and. .not. scm_state%do_sst_initialize_only)) then`

Adding the parenthesis around the `.and.` statement might help make it clearer to people quickly glancing at the code?

scrasmussen (approved):
Changes look good to me! Probably want another reviewer with a better understanding of the physics but from a coding stand point it seems good to me.

hertneky (approved):
These changes look good to me.

PR #596, opened 2025-06-23
Combination of NRL PRs and UFS/dev #304

grantfirl:
SOURCE: @grantfirl

DESCRIPTION OF CHANGES:
  - Point to combination ccpp-physics branch

ISSUE: None

ASSOCIATED PRs:
https://github.com/NCAR/ccpp-physics/pull/1159

TESTS CONDUCTED: SCM RTs and UFS RTs

REGRESSION TEST CHANGES: None expected.


grantfirl:
@hertneky @scrasmussen I merged this because it was only updating the ccpp/physics submodule (and that PR was needed by NRL) and the RTs were identical to the previous PR. We still need to update baselines (either from this PR or from the previous one -- they should be identical).

hertneky:
> @hertneky @scrasmussen I merged this because it was only updating the ccpp/physics submodule (and that PR was needed by NRL) and the RTs were identical to the previous PR. We still need to update baselines (either from this PR or from the previous one -- they should be identical).

Done

PR #602, opened 2025-07-01
hotfix: Github CI MPI Bug

scrasmussen:

DESCRIPTION OF CHANGES:
  - SCM's CI is failing, looks like it is picking up Conda's MPI rather than the system's version. Using APT to install NetCDF and OpenMPI to see if that will fix the issue. This is similar to the method of the [Dockerfile](https://github.com/NCAR/ccpp-scm/blob/main/docker/Dockerfile), which works. Not the best fix but trying to get something quick in to fix the issue.

TESTS CONDUCTED: Github CI will work if the tests pass


scrasmussen:
This PR is ready for review. The conda installation of MPI seemed to have started conflicting with the systems version, causing the error in the CMake build phase. This PR does the following

- Use APT packager manager to install NetCDF and OpenMPI
- Bumping GNU version to 13, this is the version APT installs. In the future actions will use Docker images to use different versions
- Switch to pip instead of conda

I played around with using a `dtcenter/ccpp-scm:base` Docker image to handle the dependencies of the Github actions. This will work and save us time and effort in the future, just need to figure out how to best handle setting and using different compiler versions.

dustinswales (approved):

PR #604, opened 2025-07-14
ufs/dev PR295

dustinswales:
SOURCE: Dustin Swales NOAA-GSL

DESCRIPTION OF CHANGES:
SCM changes identical to FV3UFS [NOAA-EMC/fv3atm #950](https://github.com/NOAA-EMC/fv3atm/pull/950). 

ISSUE:
N/A

ASSOCIATED PRs:
 - NCAR/ccpp-physics [#1149 ](https://github.com/NCAR/ccpp-physics/pull/1149)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,8 +4,8 @@
 	branch = main
 [submodule "ccpp-physics"]
 	path = ccpp/physics
-	url = https://github.com/dustinswales/ccpp-physics
-	branch = ufs-dev-PR295
+	url = https://github.com/NCAR/ccpp-physics
+	branch = main
 [submodule "CMakeModules"]
 	path = CMakeModules
 	url = https://github.com/noaa-emc/CMakeModules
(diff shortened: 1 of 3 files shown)


scrasmussen (approved):
Approving because you said there are fewer changes here but pointing them out just to double check, in FV3UFS [NOAA-EMC/fv3atm #950](https://github.com/NOAA-EMC/fv3atm/pull/950) we have the following class of changes that aren't in this PR
- `get_tracer_index` -> `get_physics_tracer_index`
- `integer, parameter :: physics_no_tracer = -99`

grantfirl:
The debug mode failure should be debugged at some point, but not here. It doesn't fail on any other platforms.

scrasmussen:
> The debug mode failure should be debugged at some point, but not here. It doesn't fail on any other platforms.

Agreed, I was able to reproduce this error locally which will make it easier to debug. I'll track it down.

PR #613, opened 2025-08-27
bugfix: CI fix in module_mp_nssl_2mom.F90

scrasmussen:

DESCRIPTION OF CHANGES:
  - bugfix: initialize data so it is never NaN. Fixes bug in CI when running the `arm_sgp_summer_1997_A` case and `SCM_WoFS_v0` suite. See Issue NCAR/ccpp-physics#1155 for details
  
ISSUE: NCAR/ccpp-physics#1155

ASSOCIATED PRs:
 -  NCAR/ccpp-physics#1157

TESTS CONDUCTED: CI

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,8 +4,8 @@
 	branch = main
 [submodule "ccpp-physics"]
 	path = ccpp/physics
-	url = https://github.com/scrasmussen/ccpp-physics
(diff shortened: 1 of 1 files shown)


grantfirl (approved):

grantfirl:
Thanks for the CI fix. There is also an issue in UWM for the NSSL MP debug mode, so this should probably fix that too.

PR #625, opened 2025-10-16
Create RT Artifacts on Main Branch

scrasmussen:

DESCRIPTION OF CHANGES:
  - Run SCM RTs on main branch so artifacts will be created for it. In the future these will be used to compare with instead of downloading them from the dtcenter's web server
  - Add  `-${{artifact_origin}}` name to artifacts so the artifacts on the `main` branch can easily be retrieved in the future

TESTS:
 - Once branch is merged `main` should then have artifacts. Checked the `artifact_origin` name is working correctly by locally testing with `act`

### DIFF
--- a/.github/workflows/ci_build_scm_ubuntu_nvidia.yml
+++ b/.github/workflows/ci_build_scm_ubuntu_nvidia.yml
@@ -1,6 +1,6 @@
 name: build the CCPP-SCM with Nvidia
 
-on: [pull_request,workflow_dispatch]
+on: workflow_dispatch
 
 jobs:
 


grantfirl:
@scrasmussen So, this PR only to create the main-based artifacts? This gets triggered on a push to main, which should ideally only happen once we press the merge button, right? Afterwards, there will be another PR to actually use the main artifacts, correct?

grantfirl (approved):

scrasmussen:
> @scrasmussen I changed the NVidia workflow to only run on workflow dispatch. Is this what you had in mind? This way, we can keep it in the repo and test it manually if we ever need to, but it won't run automatically during a PR.

@grantfirl This is perfect, I was going to remove it all together but keeping it as a workflow dispatch is better. Thanks for the change!



> @scrasmussen So, this PR only to create the main-based artifacts? This gets triggered on a push to main, which should ideally only happen once we press the merge button, right? Afterwards, there will be another PR to actually use the main artifacts, correct?

Exactly, correct on all accounts.

PR #628, opened 2025-10-20
Read RTs from Main Branch Artifacts

scrasmussen:

DESCRIPTION OF CHANGES:
  - Read RTs from main branch artifacts instead of the DTC's server
  - This means the artifacts won't need to be manually uploaded
  - Here a link to the [download-artifact](https://github.com/actions/download-artifact?tab=readme-ov-file#inputs) Github action 

TESTS CONDUCTED: The CI actions from the workflow will test if this works or not, a `github-token` may be needed

### DIFF
--- a/.github/workflows/ci_run_scm_rts.yml
+++ b/.github/workflows/ci_run_scm_rts.yml
@@ -33,6 +33,7 @@ jobs:
       dir_rt:     /home/runner/work/ccpp-scm/ccpp-scm/test/artifact-${{matrix.build-type}}
       dir_bl:     /home/runner/work/ccpp-scm/ccpp-scm/test/BL-${{matrix.build-type}}
       artifact_origin: ${{ github.event_name == 'pull_request' && 'PR' || ('main' == 'main' && 'main' || 'PR') }}
+      GH_TOKEN:   ${{ github.token }}
 
     # Workflow steps
     steps:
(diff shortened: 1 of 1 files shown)


scrasmussen:
This PR is being moved from draft to "ready to review". 

Reviewers can double check it is functioning correctly by going to the Github Action Checks Tab -> 
`build and run SCM regression` tests -> select a test and go to `Download SCM RT baselines` -> check `artifact_id` value matches the `main` branch action's number (it will be the first long number in that url).

grantfirl:
@hertneky @scrasmussen It looks like the comparison to baseline portion of the CI RTs was still failing after this PR. I *think* it is because using the `download-artifact` action doesn't unzip it. I've added an unzip command for the baseline artifact to the next section of the script in https://github.com/NCAR/ccpp-scm/pull/630/commits/eb825ae2aeb2dda3c843ce4e084be3ac5a4e3134. Hopefully, this will work.

grantfirl:
> @hertneky @scrasmussen It looks like the comparison to baseline portion of the CI RTs was still failing after this PR. I _think_ it is because using the `download-artifact` action doesn't unzip it. I've added an unzip command for the baseline artifact to the next section of the script in [eb825ae](https://github.com/NCAR/ccpp-scm/commit/eb825ae2aeb2dda3c843ce4e084be3ac5a4e3134). Hopefully, this will work.

@scrasmussen @hertneky Nevermind about the unzip thing. It looks like there was a different problem since the docs say that the `download-artifact` action automatically unzips. I'm debugging it in #630

PR #630, opened 2025-10-20
UFS-dev PR#305

grantfirl:
SOURCE: @grantfirl 

DESCRIPTION OF CHANGES:
  - Code changes need for RRTMGP v1.8

ISSUE: None

ASSOCIATED PRs:
https://github.com/NCAR/ccpp-physics/pull/1168
https://github.com/NCAR/ccpp-physics/pull/1172

TESTS CONDUCTED: SCM/UFS RTs

This PR catches the NCAR:main branch up with changes from the ufs-community:ufs/dev branch.

Associated ufs/dev PR:
- https://github.com/ufs-community/ccpp-physics/pull/305

Associated ufsatm PR:
- https://github.com/NOAA-EMC/ufsatm/pull/995

Associated NCAR PR:
 - https://github.com/NCAR/ccpp-physics/pull/1168

---

REGRESSION TEST CHANGES: Assuming all RRTMGP-based suites should change results.

PR #636, opened 2025-11-07
hotfix: remove gh artifact retrieval limit

scrasmussen:
SOURCE: Soren Rasmussen, NCAR

DESCRIPTION OF CHANGES:
  - CI was failing since the `gh` command only finds the past 30 artifact ids by default. Adding the `--paginate` flag lists all artifacts that exist. Note, this doesn't list old ones, just ones that can be downloaded.

ISSUE:  CI was failing since the `gh` command only finds the past 30 artifact ids by default. After enough PR CIs were run and created artifacts, the main branch artifacts were pushed off the returned list.

ASSOCIATED PRs:
 - This issue showed up in #630 

TESTS CONDUCTED: List tests done as appropriate. Delete if not used.

### DIFF
--- a/.github/workflows/ci_create_scm_rts_artifacts.yml
+++ b/.github/workflows/ci_create_scm_rts_artifacts.yml
@@ -1,0 +1,187 @@
+
+name: create scm rt artifacts
+
+on:
+  workflow_dispatch:
+
+jobs:
+  run_scm_rts:
+
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        fortran-compiler: [12]
+        build-type:       [Release, Debug, SinglePrecision]
+        py-version:       [3.11]
+
+    continue-on-error: true
+
+    # Environmental variables
+    env:
+      NFHOME:     /home/runner/netcdf-fortran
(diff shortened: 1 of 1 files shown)


scrasmussen:
@grantfirl 

> to get around this, I was thinking that we can just have a workflow-dispatchable CI script to generate a fresh artifact from main.

These are good points, sorry this new testing feature didn't initially quite work as expected! I had considered adding a workflow-dispatchable CI script, what if the artifacts had expired from being too old, there would be no way to create new ones... I'll add a new script to this PR to do that. 

 > but it errors out due to not being able to download existing artifacts before uploading newly-created ones.
> 
> I'm also wondering if we should change the triggers on the RT CIs to either only be workflow dispatchable or dispatchable via label. Having them automatically run every time there is a push to a PR is probably overkill.

I think there are pros and cons to doing it each way. Should new `main` artifacts only be pushed after they've been compared with the previous ones? Is it safer or should we consider any commit hitting the `main` branch to be truth by default, it would have gone through a vetting process already.

I prefer the overkill method, but am open to this discussion! I've added these points to our Tuesday SCM meeting notes so we can talk about them

PR #649, opened 2025-12-28
Update: Derecho Spack Stack 1.9.3 Modules

scrasmussen:

DESCRIPTION OF CHANGES:
  - Change derecho_{gnu, intel} module to be a sym-link to the newest spack stack file
  - Adding module files that use spack stack 1.9.3 

TESTS CONDUCTED: 
 - Successfully build with GNU module
 - Successfully builds with Intel module and new `ifx` compiler
   - Had to turn off `-g` and most Fortran compiler flags. Optimization `-O0` worked but not `-O2`
     - `-g` seems to turn on the Parallel MPI compiler and errors like the following appear: `ld: CMakeFiles/scm.dir/scm.F90.o:(.debug_info+0x151): undefined reference to MPIR_C_MPI_ARGVS_NULL'`
   - Users can't use a login node to build the file `ccpp_static_api.F90` because it runs out of memory and gets an Internal Compiler Error (ICE). Added a mesage letting them know they need to use an interactive node to get more memory. Successfully built with `mem=100GB`, likely a smaller amount could be used


scrasmussen:
Latest commit has the changes required to build with Intel ifx on Derecho. Not sure if this will be required on other systems, but on Derecho the login node didn't have enough memory and most compiler flags needed to be turned off for it to build. This is reflected in the CMakeLists.txt and module files.

PR #524, opened 2024-10-07
Use ip@5: if available instead of sp

climbfuji:
The sp library is deprecated and replaced by ip@5 and newer (drop-in replacement). While sp@2.5.0 is still in spack-stack-1.8.0, ip-5.0.0 is also available. We expect sp to be removed in spack-stack-1.9.0.

This PR goes with https://github.com/NCAR/ccpp-physics/pull/1090

PR #662, opened 2026-02-26
Update: spack stack 2.0.0 (for Intel) and 2.1.0 (for GNU) and ip library

scrasmussen:

DESCRIPTION OF CHANGES:
  - Add ability to build with `ip` library if it is found. The `sp` library is being replaced by `ip` so this is required. Note that in spack-stack the `ip` package [builds with the OpenMP flag](https://github.com/NOAA-EMC/NCEPLIBS-ip/blob/bc0048f25ee4fee9baf0a3cb21a3ff2abe62c4fb/spack/package.py#L88) so it gets turned on in the CMake build. The RRTMGP files currently break if compiled with OpenMP flags so the `CMAKE_Fortran_FLAGS_OPENMP_OFF` environment variable will make sure OpenMP isn't used by the RRTMGP files.
  - Bacio target was updated from `bacio::bacio_4` to `bacio::bacio` in spack-stack 2.0.0
  - To build with the Intel oneAPI 2025.2.1 the optimization had to be turned off to `-O0`. It fails at runtime on Derecho though, this will need to investigated more. 
  - Spack Stack 1.9.3 is used for GNU and 2.0.0 for Intel since [per the Spack Stack 2.0.0 release notes](https://github.com/JCSDA/spack-stack/releases/tag/spack-stack-2.0.0), GNU is not supported on Derecho in 2.0.0

ASSOCIATED PRs:
 - NCAR/ccpp-physics#1197
 - https://github.com/NCAR/ccpp-physics/pull/1208

TESTS CONDUCTED: Built on Derecho with GNU and Intel. GNU successfully ran the scm cases but Intel is failing. 

## Contributors:
Built on work from @climbfuji in NCAR/ccpp-physics#1090 and NCAR/ccpp-scm#524


scrasmussen:
> Did you ever discover what the issue with running the SCM was for Intel on Derecho? Do you want me to look into it? LMK

@hertneky I haven't had time to dive into why the SCM is failing with an Intel executable. I'll be able to look into it a bit more next week. If we want to merge this in sooner I can split off the intel stuff until we can get it to run. Or we can add a note when the spack stack 2.0 intel modules are loaded that the SCM is failing at runtime with them

scrasmussen:
I believe this PR is now ready to be merged after people review it.

Things I updated and tested today:
 - Looks like the Intel built executable was failing at runtime on Derecho because `SCM_GFS_v16_ps` is a required suite to run the supported case `arm_sgp_summer_1997_A`. I updated the supported suites to reflect that need
 - The code builds and runs for `derecho_gnu/2.1.0` and `derecho_intel/2.0.0`
   - `spack-stack 2.1.0` for Intel doesn't currently work on Derecho because it tries to load `intel/2025.3.1` and only `intel/2025.3.2` exists
   - Just noting that sadly the `emacs` module doesn't exist with `spack-stack 2.1.0`
 - I've add `derecho_gnu` and `derecho_intel` as directories and the different spack-stack versions as lua files in them. The default version loaded is listed in the ` .modulerc.lua` files

scm/src/supported_suites.py
```diff
@@ -1,2 +1 @@
-suites = ["SCM_GFS_v16","SCM_GFS_v16_RRTMGP","SCM_GFS_v17_p8_ugwpv1","SCM_HRRR_gf","SCM_WoFS_v0"]
-
+suites = ["SCM_GFS_v16","SCM_GFS_v16_ps","SCM_GFS_v16_RRTMGP","SCM_GFS_v17_p8_ugwpv1","SCM_HRRR_gf","SCM_WoFS_v0"]
```
grantfirl: I don't think that adding SCM_GFS_v16_ps should be necessary. Let me look.

grantfirl:
FYI, this now depends on https://github.com/NCAR/ccpp-physics/pull/1208

scm/src/supported_suites.py
grantfirl: I don't think that supported_suites and supported_cases are actually used anymore by anything. I really don't understand why this was necessary.

scrasmussen: I and I believe @hertneky, use supported suites and cases as the minimum list of suites and cases to test. I didn't use them before `ifx`, because I'd just use Intel to build everything at once.

I agree they don't seem to be used by anything, so if they don't represent the minimum list of suites and cases we support, we might want to discuss removing these files in the SCM meeting.

grantfirl: Ya, I think these have been entirely supplanted by suite_info.py. I think that they should be removed.

grantfirl:
@scrasmussen There will be an incoming PR into your PR branch to add functionality to compile suites based on the `run_list` functionality in the rt_test_cases.py. Compiling only those suites needed to run each of the run_lists in rt_test_cases.py at one time allows one to get past the ifx memory issue easily without having to manually specify suites for cmake.

grantfirl:
@scrasmussen Please see https://github.com/scrasmussen/ccpp-scm/pull/11

I uses this successfully on derecho using intel and gnu. The only issue that I ran into was the interactive session didn't have enough memory for the RRTMGP runs by default.

scm/src/supported_suites.py
scrasmussen: Ok, I see now, thanks for the explanation! I'll add a commit removing those two files

hertneky:
@scrasmussen I had no issues building and running on ursa and derecho now. Thank you for switching back to building supported by default. Lastly, we should update the documentation for spack-stack as part of this PR. Do you want me to do this?
I also notice that the README doesn't list the SCM-latest UG anymore - I thought it used to, but maybe I am remembering wrong.

scrasmussen:
Glad it’s working now! I can update the documentation, good thinking. I’m
on PTO today but will get that done on Monday.

scrasmussen:
@grantfirl @hertneky 
I updated the documentation with the changes we discussed yesterday. I believe this PR is good to go so feel free to review as time permits. Thanks!

grantfirl (approved):

PR #678, opened 2026-03-24
suite_info fix

scrasmussen:

DESCRIPTION OF CHANGES:
  - default regression tests in suite_info.py throw an error during build process
  - passing compiler to suite_info.py allows it to choose the subset of tests that will work
```
Traceback (most recent call last):
  File "/glade/derecho/scratch/soren/src/ccpp/prs/prxy4-ci-hotfix/scm/src/suite_info.py", line 124, in <module>
    main()
  File "/glade/derecho/scratch/soren/src/ccpp/prs/prxy4-ci-hotfix/scm/src/suite_info.py", line 95, in main
    import rt_test_cases_sp
ModuleNotFoundError: No module named 'rt_test_cases_sp'
```

ASSOCIATED PRs: should pass CI after PR #677 is merged

### DIFF
--- a/scm/src/suite_info.py
+++ b/scm/src/suite_info.py
@@ -1,14 +1,14 @@
 #!/usr/bin/env python
 
-import sys, os
+import sys, os, argparse
 
 #DEFAULT_SUITE_BEHAVIOR = 'supported'
 DEFAULT_SUITE_BEHAVIOR = 'regression_test'
 
 class suite(object):
-  
+
     DEFAULT_MAX_TIMESTEP = 1800.0
-  
+
     def __init__(self, name, tracers, namelist, timestep, max_timestep, supported):
         DEFAULT_MAX_TIMESTEP = 1800.0
         self._name = name                     #should remain unchanged after init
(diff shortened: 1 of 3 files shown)


scm/src/suite_info.py
```diff
@@ -81,45 +81,54 @@ def timestep(self, value):
...
         import rt_test_cases
-        import rt_test_cases_sp
-        import rt_test_cases_nvidia
-        
-        for item in rt_test_cases.run_list_supported:
-            rt_suite_list.append(item.get("suite"))
 
-        for item in rt_test_cases.run_list_legacy:
+        # choose correct compiler
+        if 'gfortran' in args.fc.lower():
```
grantfirl: I think that this needs to be 'gnu'

scm/src/suite_info.py
```diff
@@ -81,45 +81,54 @@ def timestep(self, value):
...
+            rttc = rt_test_cases.gnu_test_cases
+        elif 'intel' in args.fc.lower():
```
grantfirl, line 103: should we test for 'intelllvm' and 'intel' separately, or should just 'intel' suffice due to 'in' being able to handle substrings?

grantfirl (changes requested):

scm/src/suite_info.py
scrasmussen: Thanks! You are correct, this change is needed when I switched to `CMAKE_Fortran_COMPILER_ID` from `CMAKE_Fortran_COMPILER`

scm/src/suite_info.py
scrasmussen: If the tests were different between the old and new Intel compilers then we'd want to handle them separately, but my understanding is the tests are the same so checking for the `intel` substring works for both

grantfirl (approved):

PR #685, opened 2026-04-21
Ufs-dev pr350

hertneky:
This PR catches the NCAR:main branch up with changes from the ufs-community:ufs/dev branch. SCM RTs passed for ursa_gnu

Associated ufs/dev PR:
- ufs-community/ccpp-physics#350

Associated ufsatm PR:
- NOAA-EMC/ufsatm#1056

Associated NCAR PR:
 - NCAR/ccpp-physics#1211

---

REGRESSION TEST CHANGES:

### DIFF
--- a/ccpp/suites/suite_SCM_GFS_v17_p8_ps.xml
+++ b/ccpp/suites/suite_SCM_GFS_v17_p8_ps.xml
@@ -33,6 +33,7 @@
       <scheme>GFS_suite_interstitial_2</scheme>
       <scheme>GFS_PBL_generic_pre</scheme>
       <scheme>satmedmfvdifq</scheme>
+      <scheme>canopy_driver</scheme>
       <scheme>GFS_PBL_generic_post</scheme>
       <scheme>GFS_GWD_generic_pre</scheme>
       <scheme>unified_ugwp</scheme>
@@ -48,7 +49,6 @@
       <scheme>GFS_suite_interstitial_3</scheme>
       <scheme>GFS_DCNV_generic_pre</scheme>
       <scheme>samfdeepcnv</scheme>
-      <scheme>canopy_driver</scheme>
       <scheme>GFS_DCNV_generic_post</scheme>
       <scheme>GFS_SCNV_generic_pre</scheme>
       <scheme>samfshalcnv</scheme>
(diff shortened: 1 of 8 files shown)


ccpp/suites/suite_SCM_GFS_v17_p8_ps.xml
```diff
@@ -48,6 +48,7 @@
       <scheme>GFS_suite_interstitial_3</scheme>
       <scheme>GFS_DCNV_generic_pre</scheme>
       <scheme>samfdeepcnv</scheme>
+      <scheme>canopy_driver</scheme>
```
grantfirl: Shouldn't this be after the PBL scheme?

hertneky: absolutely - fixed

hertneky:
@grantfirl Sorry for the delay - finally got around to fixing the submodule pointer. This is ready now.

scrasmussen (approved):
Looks good to me!

grantfirl (approved):

grantfirl:
RT baselines changed for gfortran/Release/supported, but they are insignificant and near precision level. Not sure why it happened, but it shouldn't be a problem.

grantfirl:
> RT baselines changed for gfortran/Release/supported, but they are insignificant and near precision level. Not sure why it happened, but it shouldn't be a problem.

Note that this is the second PR in a row where we got baseline differences that appear to be compiler optimization related with our RT setup. I.e. Debug tests are b4b whereas Release tests are not and have miniscule differences. I hope this doesn't become a theme...

dustinswales:
> > RT baselines changed for gfortran/Release/supported, but they are insignificant and near precision level. Not sure why it happened, but it shouldn't be a problem.
> 
> Note that this is the second PR in a row where we got baseline differences that appear to be compiler optimization related with our RT setup. I.e. Debug tests are b4b whereas Release tests are not and have miniscule differences. I hope this doesn't become a theme...

It seems to be a GNU issue, which is less tested across SDFs in the UFS. 

I don't think GFS_v16 physics with GNU in release mode is ever tested in the UWM. I just went through tests/rt.conf in the UWM, and only one GNU test uses GFS_v16, [control_stochy_debug](https://github.com/ufs-community/ufs-weather-model/blob/develop/tests/rt.conf#L394), which is a debug test, and should pass w/ lower optimazation

I know that GFS_v16_RRTMGP is not tested with GNU at all.

PR #698, opened 2026-06-02
fix: scm run script works on Derecho login node

scrasmussen:
SOURCE: Soren Rasmussen, NSF

DESCRIPTION OF CHANGES:
  - Updating run script to work from Derecho login node, updating documentation with reasoning
  - The ``run_scm.py`` script calls ``mpirun -np 1 ./scm``, but ``mpirun/mpiexec``
is unavailable on some host machine's login nodes. For example Derecho
defaults to call ``./scm``, but calling an MPI compiled executable without
``mpirun/mpiexec`` is not supported by all MPI implementations, so this is not
the default on all systems.

TESTS CONDUCTED: tested on Derecho login node

### DIFF
--- a/scm/doc/TechGuide/chap_quick.rst
+++ b/scm/doc/TechGuide/chap_quick.rst
@@ -528,6 +528,12 @@ of tests with the following command:
 
   $ ./run_scm.py -f ../../test/rt_test_cases.py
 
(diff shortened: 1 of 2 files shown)


hertneky (approved):
ran quick test on derecho - all works

PR #711, opened 2026-07-29
ufs-dev PR#353 tempo version 3 and ufs-dev PR#369 main sync

scrasmussen:
This PR catches the NCAR:main branch up with changes from the ufs-community:ufs/dev branch.

Associated ufs/dev PR:
- ufs-community/ccpp-physics#353
- ufs-community/ccpp-physics#369 
  - this had `is_init_sw_gas_optics, is_init_lw_gas_optics, is_init_sw_cloud_optics, is_init_lw_cloud_optics` additions that needed to be brought over to SCM

Associated ufsatm PR:
- NOAA-EMC/ufsatm#1063

Associated ufs-weather-model PR:
- ufs-community/ufs-weather-model#3078

Associated NCAR PR:
 - NCAR/ccpp-physics#1231

Additional Changes:
 - This [commit](https://github.com/NCAR/ccpp-scm/pull/711/changes/fb0b8775fb194eb383a3689e1f61d7937d2e7c83) fixes some bugs in the data-retrieval scripts

### DIFF
--- a/contrib/get_tempo_data.sh
+++ b/contrib/get_tempo_data.sh
@@ -44,8 +44,26 @@ fi
 BASEDIR=$MYDIR/..
 
 # Change to directory containing the physics input data, download and extract archive
+mkdir -p $BASEDIR/scm/data/physics_input_data/
 cd $BASEDIR/scm/data/physics_input_data/
 wget ${verbose} https://github.com/NCAR/ccpp-scm/releases/download/v7.0.0/tempo_data.tar.gz
 tar -xvzf tempo_data.tar.gz
 rm -f tempo_data.tar.gz
+
+# The archive uses legacy MP_TEMPO_*.dat names; tempo_v3's Fortran source
+# (module_mp_tempo_cfgs.F90) hardcodes different filenames with no
+# host/namelist override, so symlink to what it actually looks for.
+#
+# qr_acr_qg_data_tempo_v3 -> HAILAWARE variant: workaround for an upstream
+# bug (see claude_bug_report.md) where initialize_arrays_qr_acr_qg() always
+# allocates the hail-aware-sized (nrhg=9) table regardless of hailaware_flag.
+ln -sf MP_TEMPO_HAILAWARE_QRacrQG.dat qr_acr_qg_data_tempo_v3
+ln -sf MP_TEMPO_QRacrQS.dat           qr_acr_qs_data_tempo_v3
+ln -sf MP_TEMPO_freezeH2O.dat         freeze_water_data_tempo_v3
+
+# ccn_activate.bin isn't in the downloaded archive at all -- it's small
+# (~35KB) and checked directly into the TEMPO submodule, so symlink it in
+# from there instead.
+ln -sf $BASEDIR/ccpp/physics/physics/MP/TEMPO/tempo_v3/tables/ccn_activate.bin ccn_activate.bin
+
 cd $BASEDIR/
--- a/contrib/get_tempo_data.sh
+++ b/contrib/get_tempo_data.sh
@@ -10,7 +10,7 @@ print_help() {
     echo "    --help           Show this help message and exit."
 }
 
-verbose="-q"
+verbose="-v"
 # Parse command-line arguments
 while [[ "$#" -gt 0 ]]; do
     case $1 in
--- a/contrib/get_tempo_data.sh
+++ b/contrib/get_tempo_data.sh
@@ -11,6 +11,12 @@ print_help() {
 }
 
 verbose="-v"
+wget_options=(
+    --tries=10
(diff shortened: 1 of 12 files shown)


scrasmussen:
Converted PR into draft because Tempo uses variable `do_sat_adj` which is missing from the SCM.

The following variable is missing from the SCM's `CCPP_typedefs.meta` even though it does exists in [ufsatm's CCPP_typedefs.F90](https://github.com/NOAA-EMC/ufsatm/blob/d159f93483c6cb8b92b41b5c28463c8a6eb45905/ccpp/data/CCPP_typedefs.meta#L3085-L3090)
```
[do_sat_adj]
  standard_name = flag_for_saturation_adjustment_for_microphysics_in_dynamics
  long_name = flag for saturation adjustment for microphysics in dynamics
  units = none
  dimensions = ()
  type = logical
```

This is part of the GFDL_interstitial_type and I'm not sure how to handle it. Will bring this up in meetings and discuss.

scm/src/GFS_typedefs.F90
```diff
@@ -1076,8 +1080,10 @@ module GFS_typedefs
     real(kind=kind_phys) :: dt_inner        !< time step for the inner loop in s
     logical              :: sedi_semi       !< flag for semi Lagrangian sedi of rain
     integer              :: decfl           !< deformed CFL factor
-    type(ty_tempo_cfg)   :: tempo_cfg       !< Thompson MP configuration information.
+    logical              :: do_sat_adj      !< flag for saturation adjustment for microphysics in dynamics
```
dustinswales, line 1083: We don't need to do_sat_adj flag in the SCM. Correct @grantfirl?

grantfirl: The SCM never calls "fast physics" MP, but TEMPO expects that this variable exists, so I guess we need it in the SCM. @scrasmussen I don't think that we should expose it to the nml, however, since setting it to true wouldn't do anything (and be misleading to a user). I'd simply set it to false in control_initialize and be done with it.

scrasmussen: @grantfirl @dustinswales thanks for looking into this and the recommendation! Just committed the change

contrib/get_tempo_data.sh
```diff
@@ -44,8 +44,26 @@ fi
...
 rm -f tempo_data.tar.gz
+
+# The archive uses legacy MP_TEMPO_*.dat names; tempo_v3's Fortran source
+# (module_mp_tempo_cfgs.F90) hardcodes different filenames with no
+# host/namelist override, so symlink to what it actually looks for.
+#
+# qr_acr_qg_data_tempo_v3 -> HAILAWARE variant: workaround for an upstream
+# bug (see claude_bug_report.md) where initialize_arrays_qr_acr_qg() always
+# allocates the hail-aware-sized (nrhg=9) table regardless of hailaware_flag.
+ln -sf MP_TEMPO_HAILAWARE_QRacrQG.dat qr_acr_qg_data_tempo_v3
```
grantfirl: @scrasmussen Are we sure that the files are just renamed? I thought that the UWM PR (https://github.com/ufs-community/ufs-weather-model/pull/3078) required new TEMPO v3 data files, so we may actually need to just update the files in the release asset and give them the proper names there.

grantfirl: @scrasmussen I've retrieved the new v3 files from Ursa and packaged them (with the correct filenames, so no linking/renaming required) into a v7.0.0 release asset under the tempo_v3_data.tar.gz filename.

scrasmussen: @grantfirl  Thanks! I was just trying to locate the correct files so thanks for doing that. The RTs are failing now but that should be a simple fix, I can also look into it

grantfirl: @scrasmussen I'm still not sure that they are actually different, but I do seem to recall Anders saying that the tables were recalculated/different in v3. The sizes were identical, though.

The RTs are now erroring out getting different v7 assets, it looks like. I'm guessing that this has to do with me uploading a new asset in that space? I'm hoping it's just something temporary. I was planning on trying to rerun the RTs in a bit to check.

grantfirl:
@scrasmussen With all of the problems downloading data from GitHub's own servers, I wonder if we should revisit the discussion in https://github.com/NCAR/ccpp-scm/pull/658 about potentially including the data in the dockerhub images?

grantfirl:
@scrasmussen All RTs finished after restarting the failed ones and it looks like the only tests that are different is the one TEMPO test, as expected.

grantfirl (approved):

grantfirl:
@scrasmussen Thanks for adding some robustness to the data downloads!

PR #717, opened 2026-08-27
Use 2 timelevels for SCM state variables and remove the leapfrog scheme code

### LINKED ISSUE #184: Separate memory for statein and stateout variables
Currently, both "statein" and "stateout" state variables are pointers pointing to the same space in memory for the forward Euler scheme. For both stochastic physics and the calculation of non-physics tendencies, it would make sense to allocate separate memory for these.

grantfirl:
SOURCE: @grantfirl CIRA

DESCRIPTION OF CHANGES:
  - Use state_X(timelevel=2) to hold the physics-updated state (gt0, gq0, etc) and state_X(timelevel=1) to hold physics-initial state (tgrs, qgrs). Set state_X(1) = state_X(2) at the end of the timestep. Note that this is now more similar to how UFS/FV3 handles the states. It wasn't previously necessary to have 2 different states, but the tendency update physics changes combined with physics schemes that handle prognostic tracers internally necessitate this change.
  - Remove the leapfrog scheme since it is no longer needed (hasnt been used since the transition to the DEPHY input format). This is just a cleanup and to avoid future maintenance and confusion.
  - Add `only` to module use statements in SCM code for better readability etc.

ISSUE: Fixes https://github.com/NCAR/ccpp-scm/issues/184

ASSOCIATED PRs:
N/A

TESTS CONDUCTED: SCM RTs

REGRESSION TEST CHANGES: Using 2 timelevels to hold these different states causes machine-precision-level differences.


lisa-bengtsson:
I just wanted to let you know that I tried this for the prognostic tracers I had included (updraft, area fraction), and it works. Thank you.

grantfirl:
> I just wanted to let you know that I tried this for the prognostic tracers I had included (updraft, area fraction), and it works. Thank you.

@lisa-bengtsson Great! Thanks for following up.

scrasmussen (approved):
Changes look good! Like the `use foo, only: bar` changes. Looked through a number of the CI artifact plots that the CI said differed, but couldn't see any change from the baseline.

hertneky (approved):
Changes look good. I made comments about removal of some leftover mentions of leapfrog in pr718 on accident - mixing up the PRS! I decided to not make the same comments here.
Tested on derecho gnu: built and ran scm rts fine.

grantfirl:
> Changes look good. I made comments about removal of some leftover mentions of leapfrog in pr718 on accident - mixing up the PRS! I decided to not make the same comments here. Tested on derecho gnu: built and ran scm rts fine.

Thanks for the review. I addressed the comments in https://github.com/NCAR/ccpp-scm/pull/718 related to the removal of the leapfrog scheme in this PR. The other comment will be addressed in that PR once we merge this one and update 718.

PR #718, opened 2026-08-28
Remove original proprietary input format

grantfirl:
SOURCE: @grantfirl CIRA

DESCRIPTION OF CHANGES:
  - Builds off of https://github.com/NCAR/ccpp-scm/pull/717
  - Removes all code related to reading in the original input format; removes any reference to the original input format

ISSUE: N/A

ASSOCIATED PRs:
This should get merged after https://github.com/NCAR/ccpp-scm/pull/717

TESTS CONDUCTED: SCM RTs

REGRESSION TEST CHANGES: None expected over and above https://github.com/NCAR/ccpp-scm/pull/717


scrasmussen (changes requested):
Things look good to me! Though, I think one thing that will need to be changed is also removing `input_type = 1` from the `dephy_converter.py`, [line 2465](https://github.com/grantfirl/ccpp-scm/blob/443bb9d35722903b49dc60bcb5e00c77d48ad64f/scm/etc/scripts/dephy_converter.py#L2465)?  

The [mod_forcing.py](https://github.com/grantfirl/ccpp-scm/blob/443bb9d35722903b49dc60bcb5e00c77d48ad64f/scm/etc/scripts/mod_forcing.py#L183-L193) script also still has `input_type` statements. Is this script still useful? If so it should probably be cleaned up, if not it could be deleted.

scm/src/scm_forcing.F90
hertneky: Remove comment?

scm/src/scm_forcing.F90
```diff
@@ -64,867 +64,470 @@ subroutine interpolate_forcing(scm_input, scm_state, in_spinup)
...
-  end do
-  !> @}
 
-end subroutine apply_forcing_leapfrog
+  end if !model time exceeds forcing time
+
+  !> @}
+end subroutine interpolate_forcing
 
 subroutine apply_forcing_forward_Euler(scm_state, in_spinup)
```
hertneky: Is this subroutine meant to be kept even though only apply_forcing_DEPHY is used now?

scm/src/scm_time_integration.F90
```diff
@@ -30,32 +30,6 @@ module scm_time_integration
...
-  scm_state%state_T(:,:,1) = (1.0 - scm_state%c_filter)*scm_state%temp_T(:,:,2) + &
-    0.5*scm_state%c_filter*(scm_state%state_T(:,:,2) + scm_state%temp_T(:,:,1))
-  scm_state%state_u(:,:,1) = (1.0 - scm_state%c_filter)*scm_state%temp_u(:,:,2) + &
-    0.5*scm_state%c_filter*(scm_state%state_u(:,:,2) + scm_state%temp_u(:,:,1))
-  scm_state%state_v(:,:,1) = (1.0 - scm_state%c_filter)*scm_state%temp_v(:,:,2) + &
-    0.5*scm_state%c_filter*(scm_state%state_v(:,:,2) + scm_state%temp_v(:,:,1))
-
-end subroutine
-
 !> This subroutine calls nuopc_rad_update and nuopc_rad_run in nuopc_physics.F90 (if necessary) and apply_forcing_leapfrog from \ref forcing and nuopc_phys_run, also from nuopc_physics.F90.
```
hertneky: Can we remove mention of leapfrog in these comments - this one and L35.

scm/src/scm_time_integration.F90
```diff
@@ -73,35 +47,7 @@ subroutine do_time_step(scm_state, physics, in_spinup)
   !! @{
 
   !> - Call apply_forcing_* from \ref forcing. This routine updates the "input" state variables for the physics call (updates filtered values from previous timestep, if leapfrog scheme). It effectively replaces the change of the state variables due to dynamics.
```
hertneky, line 46: also mention of leapfrog here

scm/src/scm_input.F90
hertneky: sorry if i'm being too picky - it looks like the hardcoded profile was removed so this comment can remove '(temporarily hard-coded profile)'?

hertneky (changes requested):
@grantfirl Changes look good. Just a handful of comments on possible additional changes.
Builds - check!
Runs SCM RTs - check!

scm/src/scm_forcing.F90
grantfirl: Addressed in https://github.com/NCAR/ccpp-scm/pull/717

scm/src/scm_time_integration.F90
grantfirl: Addressed in https://github.com/NCAR/ccpp-scm/pull/717

scm/src/scm_time_integration.F90
grantfirl: Addressed in https://github.com/NCAR/ccpp-scm/pull/717

scm/src/scm_input.F90
grantfirl: Addressed in https://github.com/NCAR/ccpp-scm/pull/717

scm/src/scm_forcing.F90
grantfirl: No, good catch. It has been removed in https://github.com/NCAR/ccpp-scm/pull/718/commits/6ee87c3b0de8361aa006f957ac5526a1b191b007

hertneky (approved):

grantfirl:
> Things look good to me! Though, I think one thing that will need to be changed is also removing `input_type = 1` from the `dephy_converter.py`, [line 2465](https://github.com/grantfirl/ccpp-scm/blob/443bb9d35722903b49dc60bcb5e00c77d48ad64f/scm/etc/scripts/dephy_converter.py#L2465)?
> 
> The [mod_forcing.py](https://github.com/grantfirl/ccpp-scm/blob/443bb9d35722903b49dc60bcb5e00c77d48ad64f/scm/etc/scripts/mod_forcing.py#L183-L193) script also still has `input_type` statements. Is this script still useful? If so it should probably be cleaned up, if not it could be deleted.

I fixed dephy_converter.py to remove the input_type from the newly-converted case configuration files. I'm not sure if anyone is still using this script, but at least this won't be a continuing problem if someone is using this to convert old format cases out there.

I'm not really sure what to do with the mod_forcing.py script. I *think* I wrote this in order to add selected tendencies to the forcing terms so that a subsequent run of a case could remove one or more schemes from a SDF thereby removing interactions with the remaining physics (to more effectively isolate result changes from the remaining part of the suite). I doubt that this has ever been used since it was put in the repo, yet I'm kinda hesitant to remove it since the idea behind it could still be useful. I don't want to spend the time updating it, yet I don't want to remove it. If you guys prefer to remove it, I'll go along with it because it should be retrievable via git history if anyone ever needs it (or recreated altogether). Thoughts?

grantfirl:
@scrasmussen This brings up a broader point about files in scm/etc/scripts. This has kinda been a dumping ground for scripts that were useful at one point but are likely non-working in various ways and are unsupported. I think that the idea of keeping them in the repo is just that "these could be useful for someone someday, even if they're broken". We've never really supported much in here besides the UFS_case_gen script. Keeping old scripts in here probably goes against good repository "hygiene", yet I still don't know whether removing them is the best thing to do or not.

scrasmussen:
> @scrasmussen This brings up a broader point about files in scm/etc/scripts. This has kinda been a dumping ground for scripts that were useful at one point but are likely non-working in various ways and are unsupported. I think that the idea of keeping them in the repo is just that "these could be useful for someone someday, even if they're broken". We've never really supported much in here besides the UFS_case_gen script. Keeping old scripts in here probably goes against good repository "hygiene", yet I still don't know whether removing them is the best thing to do or not.

@grantfirl I like the idea of moving the old scripts we don't use anymore to a new directory named `scm/etc/scripts/archive`, or something similar. It will serve the purpose of cleaning up the `scripts` directory while still keeping the scripts around for reference.

scrasmussen (approved):
The changes look good to me! I don't mind waiting to clean up the script directory at a later point.

PR #719, opened 2026-09-05
Add CMake Source Directory Guard and Spack Stack 2.1.1 Module Files

scrasmussen:
SOURCE: Soren Rasmussen

## DESCRIPTION OF CHANGES
  - CMake source directory has previously been moved from `scm/src/` to the top of the directory. Add error message if user's CMake configure option is pointing to the old directory. This will help users get used to the change
  - Adding 2.1.1 spack stack module for GNU and Intel and updating GNU 2.1.0 to work after the Derecho system update
 
## TESTS CONDUCTED
Built SCM_HRRR_gf suite and ran twpice case for Spack  on Derecho. Ursa builds but remains untested.

| Passing Case | Suite       | Namelist          | Status |
| --- | --- | --- |--- |
| twpice | SCM_HRRR_gf | input_HRRR_gf.nml | 0      |

## NOTE
For the Ursa modules I commented out the following line because the equivalent path does not exist for Spack Stack 2.1.1
``` lmod
-- execute{cmd="source /scratch3/BMC/gmtb/ccpp-scm-software/spack-stack-2.0.0-gnu/bin/activate", modeA={"load"}}
```

### DIFF
--- a/scm/etc/modules/derecho_gnu/.modulerc.lua
+++ b/scm/etc/modules/derecho_gnu/.modulerc.lua
@@ -1,1 +1,1 @@
-module_version("derecho_gnu/2.1.0", "default")
+module_version("derecho_gnu/2.1.1", "default")
--- a/scm/etc/modules/derecho_gnu/2.1.1.lua
+++ b/scm/etc/modules/derecho_gnu/2.1.1.lua
@@ -6,10 +6,9 @@ the CISL machine Derecho (Cray) using GNU 13.3.1
 whatis([===[Loads spack-stack libraries needed for building the CCPP SCM on Derecho with GNU compilers]===])
 
 
-local spack_root = "/glade/work/epicufsrt/contrib/spack-stack/derecho/spack-stack-2.1.0/envs/ue-gcc-13.3.1/"
+local spack_root = "/glade/work/epicufsrt/contrib/spack-stack/derecho/spack-stack-2.1.1/envs/ue-gcc-13.3.1/"
 prepend_path("MODULEPATH", pathJoin(spack_root, "modules/Core/"))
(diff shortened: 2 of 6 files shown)


scrasmussen:
@grantfirl 

Did you set this up for Spack Stack 2.0.0 on Ursa? How do we create this for 2.1.1?

```
execute{cmd="source /scratch3/BMC/gmtb/ccpp-scm-software/spack-stack-2.0.0-gnu/bin/activate", modeA={"load"}}
```

hertneky (approved):
Thanks for these updates @scrasmussen. The environment works again for the the derecho gnu environment. Built and ran SCM RTs perfectly on derecho and ursa (gnu).
It would be good to update that activation script before merging - I saw you already ask Grant about it. I also don't know how it is created.

grantfirl:
> @grantfirl
> 
> Did you set this up for Spack Stack 2.0.0 on Ursa? How do we create this for 2.1.1?
> 
> ```
> execute{cmd="source /scratch3/BMC/gmtb/ccpp-scm-software/spack-stack-2.0.0-gnu/bin/activate", modeA={"load"}}
> ```

@scrasmussen Sorry for not seeing this earlier. IIRC, I followed NOAA RDHPCS's guidance on setting up a custom python environment here: https://docs.rdhpcs.noaa.gov/software/python/index.html

I don't know how Derecho likes to handle custom python environments. I think that the whole point of the custom python environment for the SCM was to use the same python environment as the underlying spack-stack installation and then add whatever packages that the SCM needs on top of it, saved in that custom environment. If this can be achieved in a different way for derecho, I think that would be OK too.

grantfirl (approved):
LGTM. Feel free to merge or wait to see if the custom python env can be figured out for Derecho.

